PR #20, opened 2021-06-23
Update formik version that uses a good version of lodash

lezzago:
### Description
Fix lodash dependency where formik uses lodash 4.17.20
 
### Issues Resolved
N/A
 
### Check List
  - [X] All tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "devDependencies": {
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "^7.0.0",
+    "cypress": "^6.0.0",
     "husky": "^3.0.0",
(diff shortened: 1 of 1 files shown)


skkosuri-amzn (approved):

AWSHurneyt (approved):

package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "brace": "0.11.1",
-    "formik": "^2.2.5",
+    "formik": "^2.2.6",
```
tlfeng, line 38: Instead of changing the version number in `package.json`, I suggest running `yarn upgrade` to update all the dependencies (including formik).
I think there are 2 reasons:
1. The symbol `^` stands for using the same major version as stated of the dependency, regardless the minor version. It will update `yanr.lock` file by picking the latest version of the dependency.
2. AD dashboards pluign is using version number of `^2.2.5` (https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/package.json#L47), if we change it alone, with the dependency version consistency strategy of OpenSearch Dashboards, it will block people in building the 2 plugins together in same folder.

PR #22, opened 2021-06-30
Show Error Toast Message whenever action execution fails from backend due to incorrect configurations

aditjind:
### Description

By design we don't throw an exception whenever we are not able to send an alert. We mark that alert in the error state and send that error as a part of the response of the monitor execution result. Since the monitor execution succeeds the response doesn't throw any error and in the [frontend code](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/main/public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js#L116) we check for the response status of the request and not the status of the action (which is present in the response content). This results in the front end not showing an error for the action which was not executed.

This implementation is correct from the backend as the monitor execution succeeded and the action is marked in the error state. The problem lies in the front-end response parsing where we should check explicitly for the action response.

This PR addresses that and parses the response to specifically check
 
### Issues Resolved
https://github.com/opensearch-project/alerting/issues/100
 
### Check List
  - [X] All tests pass
  - [X] Added unit tests

Tested by providing an incorrectly configured Web-hook URL. The `Send Test Message` button shows an error toast message. 

Screen Shot 2021-06-30 at 3 38 06 PM

### DIFF
--- a/public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
+++ b/public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
@@ -98,8 +98,7 @@ class ConfigureActions extends React.Component {
     }
   };
 
-  checkForError = (response) => {
-    let error =  null;
+  checkForError = (response, error) => {
     for (const trigger_name in response.resp.trigger_results) {
       // Check for errors in the trigger response
       if (!response.resp.trigger_results[trigger_name].error) {
@@ -111,7 +110,6 @@ class ConfigureActions extends React.Component {
         error = response.resp.trigger_results[trigger_name].error
       }
     }
-    console.log("Here is the Error Response", error)
     return error
   }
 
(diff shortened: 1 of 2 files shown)


leeyun-amzn (dismissed):

public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
```diff
@@ -43,6 +43,21 @@ const createActionContext = (context, action) => ({
   },
 });
 
+export const checkForError = (response, error) => {
+  for (const trigger_name in response.resp.trigger_results) {
+    // Check for errors in the trigger response
+    if (!response.resp.trigger_results[trigger_name].error) {
+      // Check for errors in the actions configured
+      for (const action_result in response.resp.trigger_results[trigger_name].action_results) {
+        error = response.resp.trigger_results[trigger_name].action_results[action_result].error;
```
lezzago, line 52: Is it possible for 1 result to have an error and not the other? This causes the second run of the for loop to hide the error

aditjind: This is from the Test Message Button Ashish! Only single action is provided as an input. The response is rendered as a list because the response from actual monitor run can contain several actions. We are sure this will contain single action result.

leeyun-amzn (approved):

lezzago (approved):

PR #29, opened 2021-07-13
Allow for custom endpoints for cypress tests

lezzago:
### Description
Allow for custom endpoints for cypress tests
 
### Issues Resolved
N/A
 
### Check List
  - [X] All tests pass

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,6 +1,9 @@
 {
   "defaultCommandTimeout": 10000,
   "env": {
+    "base_url": "localhost",
+    "opensearch_port": "9200",
+    "opensearch_dashboards_port": "5601",
     "opensearch": "localhost:9200",
     "opensearch_dashboards": "localhost:5601",
     "security_enabled": false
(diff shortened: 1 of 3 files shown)


thalurur:
I believe the only reason to have the specific if condition for security enabled case is to ensure the use of https instead of http. Introducing new constants base_url and ports can potentially conflict with the defined opensearch variable inside the cypress config? Wouldn’t it make more sense to just set the protocol to https when security enabled but still derive the value of endpoint “opensearch” env variable and have only this input option for users to customize the endpoint for opensearch?

lezzago:
> I believe the only reason to have the specific if condition for security enabled case is to ensure the use of https instead of http. Introducing new constants base_url and ports can potentially conflict with the defined opensearch variable inside the cypress config? Wouldn’t it make more sense to just set the protocol to https when security enabled but still derive the value of endpoint from “opensearch” env variable and have only this input option for users to customize the endpoint for opensearch?

I meant to clean up the `opensearch` and `opensearch_dashboards` in cypress.json, but also you made a good point that both `opensearch` and `opensearch_dashboards` endpoints can be different as well. I will modify it based on what you suggested, so users can customize the endpoints and include the ports there as well.

thalurur (approved):

PR #73, opened 2021-08-31
Create monitor page, bucket level monitor showing bar graph 

leeyun-amzn:
### Description
Support bar graph view for bucket-level monitors.
 
###Known Issues
Data goes beyond the minimum of x axis, need to filter the data.

###Screenshots

Graph for bucket-level monitor
image
Graph for query-level monitor (no change, just showing the difference) 
image

 
### Check List
  - [ ] All tests pass


public/pages/CreateMonitor/components/VisualGraph/utils/helpers.js
```diff
@@ -80,10 +96,52 @@ export function getAnnotationData(xDomain, yDomain, thresholdValue) {
...
+  } else {
+    const buckets = _.get(response, 'aggregations.composite_agg.buckets', []);
+    return buckets
+      .map((bucket) => getXYValuesByFieldName(bucket, fieldName))
+      .filter(filterInvalidYValues);
+  }
+}
+
+// Function for aggregation type monitors to get Map of data
+export function getMapDataFromResponse(response, fieldName, groupByFields) {
```
qreshi, line 115: Just to make sure I understand the logic below correctly:

We essentially have the aggregated data in the response (which is both aggregated on the group by field but also bucketed with a date histogram for the graph data series). So for each bucket in this response, we collect the metric from the bucket and save that as a data point with a key based on the bucket key values based on the group by fields (so that multiple entries of the same group by key values that can appear in different date histogram buckets are collected together). We then sort these collected results based on the metric data count that has been collected. So the top n results (based on `BAR_KEY_COUNT`) is based on the highest number of times a particular group by value tuple occurred across the time series that we're aggregating/filtering on. Is my understanding correct here?

If so, could you add a comment to this function to describe this behavior for posterity. It would help in understanding how the top values are selected and could also help if we ever decide to change how this behaves

leeyun-amzn: Your understanding is correct. Let me add it to the comment

leeyun-amzn: Updated

qreshi (approved):

PR #84, opened 2021-09-01
Update alert history graph for bucket-level monitors

leeyun-amzn:
Signed-off-by: Annie Lee <leeyun@amazon.com>

### Description
- Group alerts from 1 trigger in alert history graph. 
- Modify the tooltip for bucket-level triggers to show the count of alerts in various states at a certain time range.
- Modify legend for bucket-level monitors.
 
###  Screenshots

image
 
### Check List
  - [ ] All tests pass


public/pages/MonitorDetails/containers/MonitorHistory/MonitorHistory.js
```diff
@@ -171,7 +174,11 @@ class MonitorHistory extends PureComponent {
         },
       });
     }
-    return [...firstAlertDataPoints, ...restAlertsDataPoints, ...lastAlertDataPoint];
+
+    const triggerData = [...firstAlertDataPoints, ...restAlertsDataPoints, ...lastAlertDataPoint];
```
qreshi, line 178: Just to clarify, it looks like since we're evaluating the period of `NO_ALERTS` for the last data point based on whether `lastEndTime < windowEndTime` and since we generate the plot data from scratch each time, intermediate periods of no alerts would be lost if an Alert occurred later in time right?

For example, let's say we have a graph interval from `12:00pm - 1:00pm`, and the last Alert had an end time of `12:50pm`. In this case, we would have a 10 minute interval of `NO_ALERTS`. Then let's say at `1:01pm` there is a new Alert that is in an ongoing `ACTIVE` state. Now, when we are viewing our graph interval from `12:00pm - 1:10pm` (while that last Alert is still `ACTIVE`), will the whole graph history be red now or will we have preserved that `12:50pm-1:00pm` chunk of `NO_ALERTS`?

leeyun-amzn: So the no alert (green) state is also being pushed to `triggerData` in the helper method [dataPointsGenerator](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/aee0fc0235effe9627def18705ff5ce05e748201/public/pages/MonitorDetails/containers/MonitorHistory/utils/chartHelpers.js#L30). It checks whether there was a end time for last alert, if so push a green data point in between 2 alerts.

qreshi: Got it, makes sense.

I don't think we have to do it now, but in the future we could look into aggregating this Alert information via a date histogram of some sort so that we can have the Alerts bucketed by time intervals. Then we could represent the data as some type of layered bar graph where the different categories of Alerts in each time bucket can be visualized easily.

PR #199, opened 2022-03-29
Fix pagination issue when retrieving past the 1000th monitor mark

lezzago:
Signed-off-by: Ashish Agrawal <ashisagr@amazon.com>

### Description
The monitor page would not let the user see monitors past the 1000th monitor
 
### Issues Resolved
N/A
 
### Check List
  - [X] All tests pass

### DIFF
--- a/public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
+++ b/public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
@@ -148,7 +148,7 @@ export default class AlertsDashboardFlyoutComponent extends Component {
       this.setState({ tabContent: this.renderAlertsTable() });
   }
 
-  getBucketLevelGraphConditions = (trigger) => {
+  getMultipleGraphConditions = (trigger) => {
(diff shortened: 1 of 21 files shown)


public/pages/Monitors/containers/Monitors/Monitors.js
```diff
@@ -386,7 +384,7 @@ export default class Monitors extends Component {
       pageIndex: page,
       pageSize: size,
       pageSizeOptions: DEFAULT_PAGE_SIZE_OPTIONS,
-      totalItemCount: Math.min(MAX_MONITOR_COUNT, totalMonitors),
+      totalItemCount: totalMonitors,
```
qreshi, line 387: The table loads the `totalItemCount` items in memory right? Have we checked how this behaves now that this could potentially be unbounded. If we don't have a max for the `MAX_MONITOR` backend setting then a search for total Monitors could return up to `10,000` Monitors (I believe that's the natural limitation of the search we do).

Have we tried to populate a table this large and seen how it scales? It's possible this limitation was made at the time to safeguard the frontend.

qreshi (changes requested):

PR #218, opened 2022-04-18
Implemented UX support for configuring doc level monitors.

AWSHurneyt:
Signed-off-by: AWSHurneyt <hurneyt@amazon.com>

### Description
Refactored existing UX elements to support configuring doc level monitors.

### Screenshots
Visual editor UX
localhost_5601_app_home (3)

Extraction query editor UX
localhost_5601_app_home (2)

Monitor details page
localhost_5601_app_home (1)

Alerts by trigger flyout, and finding flyout
localhost_5601_app_home

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
+++ b/public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
@@ -14,6 +14,8 @@ import {
   EuiIcon,
   EuiLink,
   EuiSpacer,
+  EuiTab,
+  EuiTabs,
   EuiText,
 } from '@elastic/eui';
 import { getTime } from '../../../../pages/MonitorDetails/components/MonitorOverview/utils/getOverviewStats';
@@ -27,7 +29,6 @@ import {
   SEARCH_TYPE,
 } from '../../../../utils/constants';
 import { TRIGGER_TYPE } from '../../../../pages/CreateTrigger/containers/CreateTrigger/utils/constants';
-import { SEVERITY_OPTIONS } from '../../../../pages/CreateTrigger/containers/DefineTrigger/DefineTrigger';
 import { UNITS_OF_TIME } from '../../../../pages/CreateMonitor/components/MonitorExpressions/expressions/utils/constants';
 import { DEFAULT_WHERE_EXPRESSION_TEXT } from '../../../../pages/CreateMonitor/components/MonitorExpressions/expressions/utils/whereHelpers';
 import { backendErrorNotification } from '../../../../utils/helpers';
@@ -45,14 +46,18 @@ import { queryColumns } from '../../../../pages/Dashboard/utils/tableUtils';
 import { DEFAULT_PAGE_SIZE_OPTIONS } from '../../../../pages/Monitors/containers/Monitors/utils/constants';
 import queryString from 'query-string';
 import { MAX_ALERT_COUNT } from '../../../../pages/Dashboard/utils/constants';
+import { SEVERITY_OPTIONS } from '../../../../pages/CreateTrigger/utils/constants';
+import { TABLE_TAB_IDS } from '../../../../pages/Dashboard/components/FindingsDashboard/utils';
+import FindingsDashboard from '../../../../pages/Dashboard/containers/FindingsDashboard';
 
 export const DEFAULT_NUM_FLYOUT_ROWS = 10;
 
 export default class AlertsDashboardFlyoutComponent extends Component {
   constructor(props) {
     super(props);
-    const { location, monitor_id } = this.props;
-
+    const { location, monitors, monitor_id } = this.props;
+    const monitor = _.get(_.find(monitors, { _id: monitor_id }), '_source');
+    const monitorType = _.get(monitor, 'monitor_type', MONITOR_TYPE.QUERY_LEVEL);
     const {
       alertState,
       from,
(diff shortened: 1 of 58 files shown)


AWSHurneyt:
The unit test flow is currently failing due to outdated snapshots. `yarn run test:jest -u` is currently failing to update snapshots locally with the following error:
```
Test suite failed to run

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.
    
    ReferenceError: window is not defined
```
A solution seems to involve adding the following to the top of each test class. There are over 100 test classes at the moment, so that annotation will be added in a separate PR.
```
/**
 * @jest-environment jsdom
 */
```

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
```diff
@@ -45,14 +46,18 @@ import { queryColumns } from '../../../../pages/Dashboard/utils/tableUtils';
...
 export const DEFAULT_NUM_FLYOUT_ROWS = 10;
 
 export default class AlertsDashboardFlyoutComponent extends Component {
   constructor(props) {
     super(props);
-    const { location, monitor_id } = this.props;
-
+    const { location, monitors, monitor_id } = this.props;
+    const monitor = _.get(_.find(monitors, { _id: monitor_id }), '_source');
+    const monitorType = _.get(monitor, 'monitor_type', MONITOR_TYPE.QUERY_LEVEL);
```
lezzago, line 60: Why is this only for `QUERY_LEVEL`?

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
```diff
@@ -67,8 +72,9 @@ export default class AlertsDashboardFlyoutComponent extends Component {
       alerts: [],
       alertState: alertState,
       loading: true,
-      monitors: [],
+      monitor: monitor,
```
lezzago, line 75: We are now changing this from a list to a single value. Are there concerns for this?

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
```diff
@@ -153,7 +167,7 @@ export default class AlertsDashboardFlyoutComponent extends Component {
   };
 
   getAlerts = async () => {
-    this.setState({ ...this.state, loading: true });
+    this.setState({ loading: true });
```
lezzago, line 170: Why are we removing `...this.state`?

public/pages/CreateMonitor/containers/CreateMonitor/utils/formikToMonitor.js
```diff
@@ -28,16 +40,18 @@ export function formikToMonitor(values) {
...
       monitor_type: values.monitor_type,
+      ...monitorUiMetadata(),
     },
   };
 }
 
 export function formikToInputs(values) {
-  switch (values.searchType) {
-    case SEARCH_TYPE.CLUSTER_METRICS:
+  switch (values.monitor_type) {
```
lezzago, line 50: is there a concern to use `monitor_type` instead of `searchType`?

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
AWSHurneyt: On `line 60`, right? The 3rd parameter in the `_.get()` method is a default value to return if `monitor.monitor_type` is `undefined`.

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
AWSHurneyt: Not for the alerts flyout. It's intended to only display alerts for a single trigger; and that trigger is only associated with 1 monitor.

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
AWSHurneyt: I don't believe that bit is necessary. My understanding is that the `setState` method will update the `state` with only the key/value pairs provided in its argument. So if the `state` contains an `alerts` list and a `loading` boolean, and `setState` should only update the `loading` value when executing `this.setState({ loading: true })`.

public/pages/CreateMonitor/containers/CreateMonitor/utils/formikToMonitor.js
AWSHurneyt: I think we should be ok using `monitor_type` here. `ClusterMetrics` has its own search and monitor type, so either value could be used get the `switch` to call the appropriate method. Before doc level monitors were added, `formikToInputs` handled the values to either cluster metrics data or otherwise defaulted to the parsing logic used for bucket and query monitors. `formikToSearch` has logic for calling specific methods based on search type as well.

public/components/Flyout/flyouts/components/AlertsDashboardFlyoutComponent.js
lezzago: Ah ok thanks, I had misunderstood that

lezzago (dismissed):

getsaurabh02 (approved):

sbcd90 (dismissed):

lezzago (approved):

lezzago (approved):

getsaurabh02 (approved):

PR #375, opened 2022-12-01
[BUG] Issues in the UI above 200 destinations paging issue #195

### LINKED ISSUE #195: [BUG] Issues in the UI above 200 destinations
**Describe the bug**
I currently have more than 200 destinations in the alerting plugin and issues are noticeable : 

 - In the "Destinations" list, pagination above item 200 is broken : 
   - number of page is correct
   - if you click on a page representing an item above 200, you can quickly see the correct page but then you are redirected back to the page that is not bugged
  
- In the "Create monitor" page : 
 - if you create an action, the destination except the first 200 are not listed in the UI and you can't select them

**To Reproduce**
1. Create more than 200 destinations
2. Go to '/app/alerting#/destinations'
3. Play with the pages (1)
4. Create a monitor and a trigger
5. Start typing a destination that is not in the 200 firsts (2), it's missing

**Plugins**
Default install (everything)

**Host/Environment**
 - OS: Debian
 - Version: 10

Opensearch Dashboards 1.2.0

jovancacvetkovic:
### Description
Pagination and max destinations count is hardcoded to 200 items. The fix will allow all destinations to be loaded.
 
### Issues Resolved
#195
 
### Check List
- [ ] All tests pass

### DIFF
--- a/public/pages/CreateTrigger/components/Action/Action.js
+++ b/public/pages/CreateTrigger/components/Action/Action.js
@@ -78,7 +78,7 @@ const Action = ({
                   // Just a swap correct fields.
                   arrayHelpers.replace(index, {
                     ...action,
-                    destination_id: options[0].value,
+                    destination_id: options[0]?.value,
                   });
                 },
                 onBlur: (e, field, form) => {
@@ -96,7 +96,6 @@ const Action = ({
                   </React.Fragment>
                 ),
                 rowHeight: 45,
-                async: true,
                 isLoading: loadingDestinations,
               }}
             />
--- a/public/utils/constants.js
+++ b/public/utils/constants.js
@@ -62,7 +62,7 @@ export const MONITOR_INPUT_DETECTOR_ID = `inputs.${INPUTS_DETECTOR_ID}`;
 
 export const AD_PREVIEW_DAYS = 7;
 
-export const MAX_QUERY_RESULT_SIZE = 200;
+export const MAX_QUERY_RESULT_SIZE = 5000;
 
 export const MONITOR_GROUP_BY = 'ui_metadata.search.groupBy';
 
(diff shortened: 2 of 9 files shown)


public/pages/CreateTrigger/components/Action/Action.js
```diff
@@ -96,7 +96,6 @@ const Action = ({
                   </React.Fragment>
                 ),
                 rowHeight: 45,
-                async: true,
```
amsiglan, line 99: Doubt: Why did we remove the async?

public/utils/constants.js
```diff
@@ -62,7 +62,7 @@ export const MONITOR_INPUT_DETECTOR_ID = `inputs.${INPUTS_DETECTOR_ID}`;
 
 export const AD_PREVIEW_DAYS = 7;
 
-export const MAX_QUERY_RESULT_SIZE = 200;
+export const MAX_QUERY_RESULT_SIZE = 5000;
```
amsiglan, line 65: Changing this will update the max size for all the api calls, do we only want to change it for the channels?

public/pages/CreateTrigger/components/Action/Action.js
jovancacvetkovic: @amsiglan, async in this context means filtering will be done manually, look at https://eui.elastic.co/v34.6.0/#/forms/combo-box#async. The prop is set but there is no logic added to support this and hence the combo filter was not working.
The combo gets all the data anyway and the fastest filter here is the local filter which the combo already supports. I removed async in order to use standard combo search.

public/utils/constants.js
jovancacvetkovic: Right, I added MAX_CHANNELS_RESULT_SIZE=5000 to be used just for getting all the channels.

amsiglan (approved):

AWSHurneyt (approved):

PR #328, opened 2022-09-14
Fixed a bug that prevented the throttling settings from displaying for PER_ALERT bucket level trigger actions.

AWSHurneyt:
Signed-off-by: AWSHurneyt <hurneyt@amazon.com>

### Description
Fixed a bug that prevented the throttling settings from displaying for `PER_ALERT` bucket level trigger actions.
 
### Issues Resolved
[List any issues this PR will resolve]

#### When `PER_ALERT` is selected.
Screen Shot 2022-09-12 at 6 12 39 PM
#### When `PER_EXECUTION` is selected.
Screen Shot 2022-09-12 at 6 12 44 PM
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #397, opened 2022-12-05
Fixed a bug that prevented the throttling settings from displaying for PER_ALERT bucket level trigger actions

AWSHurneyt:
Signed-off-by: AWSHurneyt <hurneyt@amazon.com>

### Description
Manually backporting PR https://github.com/opensearch-project/alerting-dashboards-plugin/pull/328.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/public/pages/CreateTrigger/components/Action/actions/Message.js
+++ b/public/pages/CreateTrigger/components/Action/actions/Message.js
@@ -172,6 +172,7 @@ export default function Message(
   switch (monitorType) {
     case MONITOR_TYPE.BUCKET_LEVEL:
       displayActionableAlertsOptions = true;
+      displayThrottlingSettings = actionExecutionScopeId === NOTIFY_OPTIONS_VALUES.PER_ALERT;
       actionableAlertsSelections = _.get(values, actionableAlertsSelectionsPath);
       break;
     case MONITOR_TYPE.DOC_LEVEL:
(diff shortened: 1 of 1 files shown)


public/pages/CreateTrigger/components/Action/actions/Message.js
```diff
@@ -194,6 +195,8 @@ export default function Message(
     _.set(values, actionableAlertsSelectionsPath, actionableAlertsSelections);
   }
 
+  if (!displayThrottlingSettings) _.set(values, `${actionPath}.throttle_enabled`, false);
```
amsiglan, line 198: Is `_.set(values, `${actionPath}.throttle_enabled`, false);` set to true by default? or should we set it to true here in the else case?

AWSHurneyt: By default, `throttle_enabled` is [false](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/main/public/pages/CreateTrigger/utils/constants.js#L49). 

Setting `throttle_enabled` to false here helps ensure throttling settings aren't sent to the backend for action configurations that don't support throttling when the user may have previously been experimenting with configurations that do support such settings.

amsiglan: Aah, so if the user was playing with the setting for PER_ALERT and then goes to PER_EXECUTION then we should omit it in that case, right?

amsiglan (approved):

public/pages/CreateTrigger/components/Action/actions/Message.js
AWSHurneyt: For bucket level monitors, we want to offer `PER_ALERT` actions, but we do not want to offer them for `PER_EXECUTION`; however, the bug prevented throttling configurations for both selections.

Angie-Zhang (approved):

PR #1099, opened 2024-09-19
Fit and Finish UX changes Pt 2

jowg-amazon:
### Description
Second part of the fit and finish UX changes. This PR makes changes to the datasource alerts card which include:
- Removes truncating the monitor name
- Modifies the query that is used to grab alerts so that dataSourceId is only included as a field if data source is enabled. Without this change alerts were not fetched correctly when data source was disabled
- Alert links to the monitor page
 
Data source enabled
Screenshot 2024-09-18 at 6 03 45 PM

Data source disabled
Screenshot 2024-09-18 at 6 03 01 PM

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
+++ b/public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
@@ -88,8 +88,8 @@ export const DataSourceAlertsCard: React.FC<DataSourceAlertsCardProps> =  ({ get
           <EuiFlexItem grow={false}>
             <EuiText size="s" color="subdued">Monitor:</EuiText>
           </EuiFlexItem>
-          <EuiFlexItem grow={false} >
-            <EuiText size="m">{monitorName}</EuiText>
+          <EuiFlexItem grow={false} style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>
+            <EuiText size="m" style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>{monitorName}</EuiText>
           </EuiFlexItem>
         </EuiFlexGroup>
         <EuiHorizontalRule margin="xs" />
--- a/public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
+++ b/public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
@@ -10,7 +10,8 @@ import { getApplication, getClient, getNotifications, getSavedObjectsClient } fr
 import { dataSourceFilterFn, getSeverityColor, getSeverityBadgeText, getTruncatedText } from "../../utils/helpers";
 import { renderTime } from "../../pages/Dashboard/utils/tableUtils";
 import { ALERTS_NAV_ID, MONITORS_NAV_ID } from "../../../utils/constants";
-import { APP_PATH, DEFAULT_EMPTY_DATA, MONITOR_TYPE } from "../../utils/constants";
+import { APP_PATH, DEFAULT_EMPTY_DATA } from "../../utils/constants";
+import { dataSourceEnabled, getURL } from "../../pages/utils/helpers.js";
 
 export interface DataSourceAlertsCardProps {
   getDataSourceMenu?: DataSourceManagementPluginSetup['ui']['getDataSourceMenu'];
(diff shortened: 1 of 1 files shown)


public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
```diff
@@ -53,14 +54,22 @@ export const DataSourceAlertsCard: React.FC<DataSourceAlertsCardProps> =  ({ get
   const createAlertDetailsHeader = useCallback((alert) => {
     const severityColor = getSeverityColor(alert.severity);
     const triggerName = alert.trigger_name ?? DEFAULT_EMPTY_DATA;
+    const monitorUrl = `alerting#/monitors/${
```
amsiglan, line 57: I think we should drop the `alerting` part since the app id varies with and without new home page settings

public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
```diff
@@ -80,7 +89,7 @@ export const DataSourceAlertsCard: React.FC<DataSourceAlertsCardProps> =  ({ get
             <EuiText size="s" color="subdued">Monitor:</EuiText>
           </EuiFlexItem>
           <EuiFlexItem grow={false} >
-            <EuiText size="m">{getTruncatedText(monitorName)}</EuiText>
+            <EuiText size="m">{monitorName}</EuiText>
```
amsiglan, line 92: how will this impact in case the name is too long?

public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
jowg-amazon: Without the alerting part the URL is just appended to the all_overview page like this
`http://localhost:5601/w/HMDqJO/app/all_overview#/monitors/EjsSDJIBs61wKwycS6H4?&type=monitor&dataSourceId=`

With the alerting part the URL becomes this which links to the monitor details page
`http://localhost:5601/w/HMDqJO/app/alerting#/monitors/EjsSDJIBs61wKwycS6H4?&type=monitor&dataSourceId=`

This data source alerts card should only be present when workspaces is enabled which is only shown when the new home page settings is enabled?

public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
jowg-amazon: Updated the EuiPanel so that the overflow is hidden. This is what it would look like if the monitor name is too long

Screenshot 2024-09-19 at 2 10 24 PM

jowg-amazon:
Acknowledge alerts tests passed locally

```
       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  acknowledge_alerts_modal_spec.js         01:46        3        3        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        01:46        3        3        -        -        - 
    ```

public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
amsiglan: but this doesn't indicate there is more to the name? is there some tooltip to it? Would be better to have some truncation UI

jowg-amazon: Spoke to @xeniatup and decided to dynamically truncate the name according to the card width. Here is what it would look like on two different window sizes

Screenshot 2024-09-19 at 5 00 42 PM

Screenshot 2024-09-19 at 5 01 14 PM

public/components/DataSourceAlertsCard/DataSourceAlertsCard.tsx
jowg-amazon: Changed the id to `MONITORS_NAV_ID` instead of `"alerting"`

amsiglan (approved):

AWSHurneyt (approved):

PR #1142, opened 2024-10-25
[BugFix] AlertSummary still displays when agents required are not configured

000FLMS:
### Description
bug fix for alertSummary still displays when agents required are not configured in the selected datasource.

### Check List
  - [x] All tests pass

### Screenshots
data source with required agents configured
image

data source without agents
image

### DIFF
--- a/public/pages/Dashboard/containers/Dashboard.js
+++ b/public/pages/Dashboard/containers/Dashboard.js
@@ -45,6 +45,7 @@ import {
   getDataSourceId,
   appendCommentsAction,
   getIsCommentsEnabled,
+  getIsAgentConfigured
 } from '../../utils/helpers';
 import { getUseUpdatedUx } from '../../../services';
 
@@ -78,6 +79,7 @@ export default class Dashboard extends Component {
       totalTriggers: 0,
       chainedAlert: undefined,
       commentsEnabled: false,
+      isAgentConfigured: false,
     };
   }
 
@@ -102,17 +104,34 @@ export default class Dashboard extends Component {
     getIsCommentsEnabled(this.props.httpClient).then((commentsEnabled) => {
       this.setState({ commentsEnabled });
     });
+    this.getUpdatedAgentConfig();
   }
 
   componentDidUpdate(prevProps, prevState) {
     const prevQuery = getQueryObjectFromState(prevState);
     const currQuery = getQueryObjectFromState(this.state);
     if (!_.isEqual(prevQuery, currQuery)) {
+      this.getUpdatedAgentConfig();
       this.getUpdatedAlerts();
     }
     if (isDataSourceChanged(prevProps, this.props)) {
       this.dataSourceQuery = getDataSourceQueryObj();
+      this.getUpdatedAgentConfig();
       this.getUpdatedAlerts();
+    }
+  }
+
+  getUpdatedAgentConfig() {
+    const dataSourceId = getDataSourceId();
+    console.log("ds", dataSourceId);
+    if (dataSourceId){
+      getIsAgentConfigured(dataSourceId).then((isAgentConfigured) => {
+        this.setState({ isAgentConfigured });
+        console.log("UpAC", this.state.isAgentConfigured);
+      });
+    }
+    else{
+      this.setState({ isAgentConfigured: false });
     }
   }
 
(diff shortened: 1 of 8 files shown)


public/pages/Dashboard/containers/Dashboard.js
```diff
@@ -102,20 +104,37 @@ export default class Dashboard extends Component {
...
     if (isDataSourceChanged(prevProps, this.props)) {
       this.dataSourceQuery = getDataSourceQueryObj();
+      this.getUpdatedAgentConfig();
       this.getUpdatedAlerts();
     }
   }
 
+  getUpdatedAgentConfig() {
+    const dataSourceId = getDataSourceId();
+    console.log("ds", dataSourceId);
```
songkant-aws, line 126: Please remove test logs

public/pages/Dashboard/containers/Dashboard.js
```diff
@@ -102,20 +104,37 @@ export default class Dashboard extends Component {
...
+    if (dataSourceId){
+      getIsAgentConfigured(dataSourceId).then((isAgentConfigured) => {
+        this.setState({ isAgentConfigured });
+        console.log("UpAC", this.state.isAgentConfigured);
```
songkant-aws, line 130: Same as above

public/pages/utils/helpers.js
```diff
@@ -68,6 +72,18 @@ export const appendCommentsAction = (columns, httpClient) => {
...
+export async function getIsAgentConfigured(dataSourceId){
+  const assistantClient = getAssistantClient();
+  const res = await assistantClient.agentConfigExists(
+    [
+      SUMMARY_AGENT_CONFIG_ID,
+      LOG_PATTERN_SUMMARY_AGENT_CONFIG_ID,
+    ],
+    { dataSourceId: dataSourceId }
+  );
+  return res.exists;
```
songkant-aws, line 84: Can you also handle the error case to avoid throwing exception to corrupt the following rendering?

public/pages/Dashboard/containers/Dashboard.js
```diff
@@ -102,20 +104,37 @@ export default class Dashboard extends Component {
...
+      });
+    }
+    else{
+      this.setState({ isAgentConfigured: false });
+    }
```
songkant-aws, line 135: I think local data cluster has empty dataSourceId "". But we can also check its agent configuration. Can you check if it will be always false for local cluster?

public/plugin.tsx
```diff
@@ -234,6 +234,8 @@ export class AlertingPlugin implements Plugin<void, AlertingStart, AlertingSetup
     setApplication(core.application);
     setContentManagementStart(contentManagement);
     registerAlertsCard();
+    if (assistantDashboards)
```
songkant-aws, line 237: Could you check when we don't install assistant plugin case? If the assistant plugin is disabled or not installed, you will get null assistantClient. Then calling null assistantClient's API will definitely errored out. We can add some default behavior to avoid it.

public/pages/Dashboard/containers/Dashboard.js
000FLMS: Yes, the local cluster should be checked either, I've updated to check it.

public/pages/utils/helpers.js
000FLMS: Sure, have added a handler to return false instead of throwing exception

public/plugin.tsx
000FLMS: OK, I've set a default agentClient when the assistant plugin cannot be found.

public/pages/utils/helpers.js
```diff
@@ -74,14 +74,19 @@ export const appendCommentsAction = (columns, httpClient) => {
...
+      [
+        SUMMARY_AGENT_CONFIG_ID,
+        LOG_PATTERN_SUMMARY_AGENT_CONFIG_ID,
+      ],
+      { dataSourceId: dataSourceId }
+    );
+    return res.exists;
+  }
+  catch(e){
+    return false;
```
songkant-aws, line 88: Minor: If the call fails, we may need to notify the error message. Otherwise, we won't know whether it's dependency error or default value because of absence of plugin

riysaxen-amzn (approved):

PR #1176, opened 2024-12-18
fix: error toast message while configuring trigger while creating a monitor

vikhy-aws:
### Description
This PR fixes the issue of an unnecessary error popup appearing while creation of a monitor. The error popup is raised because the config index is not been before a monitor is created, but is created automatically when the monitor is created. So, the error popup is redundant when creating a monitor.

Error Popup
 
### Check List
- [x] Bug fix tested locally.
- [x] Added new unit test cases.
- [x] All tests passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.22.9",
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "9.5.4",
+    "cypress": "12.17.4",
     "husky": "^8.0.0",
     "lint-staged": "^10.2.0",
     "@types/react": "^16.14.23"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.22.9",
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "9.5.4",
+    "cypress": "12.17.4",
```
riysaxen-amzn, line 37: why are we changing cypress version?

server/services/DestinationsService.test.js
```diff
@@ -0,0 +1,186 @@
...
+    
+      expect(mockRes.ok).toHaveBeenCalledWith({
+        body: {
+          ok: true,
+          resp: "Indices will be configured when the monitor is created: [.opendistro-alerting-config]"
+        },
+      });
+    });
+    
+    test("should handle other errors", async () => {
```
riysaxen-amzn, line 160: what is this test about?

package.json
vikhy-aws: cypress version needs to match for OpenSearch-Dashboard and the plugins for the dashboards to run.

server/services/DestinationsService.test.js
vikhy-aws: To ensure that any other errors are caught, rather than breaking the functionality.

PR #1178, opened 2024-12-20
fix: error toast message while configuring trigger while creating a monitor

vikhy-aws:
### Description
This PR fixes the issue of an unnecessary error popup appearing while creation of a monitor. The error popup is raised because the config index is not been before a monitor is created, but is created automatically when the monitor is created. So, the error popup is redundant when creating a monitor.

image
  
### Check List
- [x] Bug fix tested locally.
- [x] Added new unit test cases.
- [x] All tests passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.22.9",
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "12.17.4",
+    "cypress": "9.5.4",
     "husky": "^8.0.0",
     "lint-staged": "^10.2.0",
     "@types/react": "^16.14.23"
@@ -67,6 +67,5 @@
   },
   "engines": {
     "yarn": "^1.21.1"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.22.9",
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "9.5.4",
+    "cypress": "12.17.4",
     "husky": "^8.0.0",
     "lint-staged": "^10.2.0",
     "@types/react": "^16.14.23"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.22.9",
     "@elastic/elastic-eslint-config-kibana": "link:../../packages/opensearch-eslint-config-opensearch-dashboards",
     "@elastic/eslint-import-resolver-kibana": "link:../../packages/osd-eslint-import-resolver-opensearch-dashboards",
-    "cypress": "9.5.4",
+    "cypress": "12.17.4",
```
amsiglan, line 37: why are we changing this? Have other plugins updated it as well?

public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
```diff
@@ -183,7 +183,10 @@ class ConfigureActions extends React.Component {
             description: '',
           }));
       } else {
-        backendErrorNotification(notifications, 'load', 'destinations', response.err);
+        // If the config index is not created, don't show the notification
+        if (response.totalMonitors !== 0) {
```
amsiglan, line 187: Club the else if?

server/services/DestinationsService.js
```diff
@@ -175,17 +175,24 @@ export default class DestinationsService extends MDSEnabledClientService {
...
     } catch (err) {
+      // Indices will be created when the monitor is created.
       if (isIndexNotFoundError(err)) {
         return res.ok({
-          body: { ok: false, resp: {} },
+          body: {
+            ok: false,
+            totalMonitors: 0, 
+            monitors: [],
+            message: "Config index will be created automatically when the monitor is created"
```
amsiglan, line 185: I would rather keep the schema same and return a known error message which we can check for in the frontend instead of introducing a new field `totalMonitors`

server/services/DestinationsService.test.js
```diff
@@ -0,0 +1,186 @@
...
+describe("Test DestinationsService -- getDestinations", () => {
+  let destinationsService;
+  let mockContext;
+  let mockReq;
+  let mockRes;
+  let mockClient;
+
+  beforeEach(() => {
+    mockClient = jest.fn();
+
```
amsiglan, line 12: Generally we don't add empty lines between initializations

package.json
vikhy-aws: So, the version of cypress in OpenSearch-Dashboards and the version of cypress in all the plugins should match for the frontend to run. So, set them to the same value.

public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
vikhy-aws: Oh sorry, I missed a case here. I got to display the error toast if the response is not okay and the totalMonitors is not zero. Will add that.

amsiglan (dismissed):

riysaxen-amzn (approved):

public/pages/CreateTrigger/containers/ConfigureActions/ConfigureActions.js
```diff
@@ -182,8 +182,9 @@ class ConfigureActions extends React.Component {
             type: toChannelType(destination.type),
             description: '',
           }));
-      } else {
-        backendErrorNotification(notifications, 'load', 'destinations', response.err);
+      } else if (response.totalMonitors !== 0) {
```
eirsep, line 185: Am hazy on my JS but are you intending to check `response.totalMonitor > 0`

if yes, why are we checking for >0 and then code comment says config index not created?

amsiglan (approved):

PR #1415, opened 2026-04-17
Added workspace ACL checks for Alerting and Monitor APIs

nishtham-amazon:
### Description                                                                                                                                                                                                    
                                                                                                                                                                                                                     
  Add workspace ACL authorization checks to all Alerting and Monitor server-side APIs to enforce workspace-level access control for OpenSearch Serverless (AOSS) data sources.                                       
                                                                                                                                                                                                                     
  **Changes:**                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - `MDSEnabledClientService.ts` — Added `checkWorkspaceAcl()` method that:                                                                                                                                          
    - Only runs for AOSS (`.aoss.amazonaws.com`) data source endpoints; skips for managed domains                                                                                                                    
    - Extracts the calling principal from `x-amzn-aosd-username` header and workspace ID from request state                                                                                                          
    - Delegates authorization to the workspace plugin's `authorizeWorkspace()` via the `WorkspacePluginStart` contract                                                                                               
    - Uses a local `WorkspaceAuthorizer` interface (structural typing) to avoid non-public import restrictions                                                                                                       
    - Added `setLogger()` for production-ready logging instead of `console.log`                                                                                                                                      
                                                                                                                                                                                                                     
  - `MonitorService.js` — Added `_enforceWorkspaceAcl()` helper and ACL checks to all API methods:                                                                                                                   
    - Write operations (`createMonitor`, `createWorkflow`, `deleteMonitor`, `deleteWorkflow`, `updateMonitor`, `executeMonitor`, `acknowledgeAlerts`, `acknowledgeChainedAlerts`): require `library_write`           
    - Read operations (`getMonitor`, `getMonitors`, `getWorkflow`, `searchMonitors`): require `library_write` or `library_read`                                                                                      
                                                                                                                                                                                                                     
  - `AlertService.js` — Added same ACL pattern to `getAlerts` and `getWorkflowAlerts`                                                                                                                                
                                                                                                                                                                                                                     
  - `plugin.js` — Updated `start()` to inject workspace plugin start contract and logger into all services                                                                                                           
                                                                                                                                                                                                                     
  - `opensearch_dashboards.json` — Added `workspace` to `optionalPlugins`                                                                                                                                            
                                                                                                                                                                                                                     
  ### Issues Resolved                                                                                                                                                                                                
                                                                                                                                                                                                                     
  <!-- List any issues this PR will resolve -->                                                                                                                                                                      
                                                                                                                                                                                                                     
  ### Check List                                                                                                                                                                                                     
    - [x] All tests pass                                                                                                                                                                                             
    - [x] New functionality has javadoc added                                                                                                                                                                        
                                                                                                                                                                                                                     
  For more information on following Developer Certificate of Origin and signing off your commits, please check                                                                                                       
  [here](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/server/alerting_configs.json
+++ b/server/alerting_configs.json
@@ -1,0 +1,3 @@
+{
+  "ws.acl.enforce.endpoint.patterns": [".aoss.amazonaws.com"]
+}
--- a/server/alerting_configs.json
+++ b/server/alerting_configs.json
@@ -1,3 +1,0 @@
-{
-  "ws.acl.enforce.endpoint.patterns": [".aoss.amazonaws.com"]
-}
--- a/server/services/AlertService.js
+++ b/server/services/AlertService.js
@@ -14,7 +14,17 @@ export const GET_ALERTS_SORT_FILTERS = {
 };
 
 export default class AlertService extends MDSEnabledClientService {
+  _enforceWorkspaceAcl = async (context, req, res, permissionModes) => {
+    const authorized = await this.checkWorkspaceAcl(context, req, permissionModes);
+    if (!authorized) {
+      return res.ok({ body: { ok: false, resp: 'Workspace ACL check failed: unauthorized' } });
+    }
+    return null;
+  };
+
   getAlerts = async (context, req, res) => {
+    const aclResponse = await this._enforceWorkspaceAcl(context, req, res, ['library_write', 'library_read']);
+    if (aclResponse) return aclResponse;
     const {
       from = 0,
       size = 20,
@@ -111,6 +121,8 @@ export default class AlertService extends MDSEnabledClientService {
   };
 
   getWorkflowAlerts = async (context, req, res) => {
+    const aclResponse = await this._enforceWorkspaceAcl(context, req, res, ['library_write', 'library_read']);
+    if (aclResponse) return aclResponse;
     const client = this.getClientBasedOnDataSource(context, req);
     try {
       const resp = await client('alerting.getWorkflowAlerts', req.query);
--- a/server/services/AlertService.js
+++ b/server/services/AlertService.js
@@ -14,16 +14,11 @@ export const GET_ALERTS_SORT_FILTERS = {
 };
 
 export default class AlertService extends MDSEnabledClientService {
-  _enforceWorkspaceAcl = async (context, req, res, permissionModes) => {
-    const authorized = await this.checkWorkspaceAcl(context, req, permissionModes);
-    if (!authorized) {
-      return res.ok({ body: { ok: false, resp: 'Workspace ACL check failed: unauthorized' } });
-    }
-    return null;
-  };
-
   getAlerts = async (context, req, res) => {
-    const aclResponse = await this._enforceWorkspaceAcl(context, req, res, ['library_write', 'library_read']);
+    const aclResponse = await this.enforceWorkspaceAcl(context, req, res, [
+      'library_write',
+      'library_read',
+    ]);
     if (aclResponse) return aclResponse;
     const {
       from = 0,
@@ -121,7 +116,10 @@ export default class AlertService extends MDSEnabledClientService {
   };
 
   getWorkflowAlerts = async (context, req, res) => {
-    const aclResponse = await this._enforceWorkspaceAcl(context, req, res, ['library_write', 'library_read']);
+    const aclResponse = await this.enforceWorkspaceAcl(context, req, res, [
+      'library_write',
+      'library_read',
+    ]);
     if (aclResponse) return aclResponse;
     const client = this.getClientBasedOnDataSource(context, req);
     try {
--- a/server/services/MDSEnabledClientService.ts
+++ b/server/services/MDSEnabledClientService.ts
@@ -1,16 +1,12 @@
 import { RequestHandlerContext, OpenSearchDashboardsRequest, OpenSearchDashboardsResponseFactory, ILegacyCustomClusterClient } from '../../../../src/core/server';
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const alertingConfig = require('../alerting_configs.json');
-
-const WS_ACL_ENDPOINT_PATTERNS: string[] = alertingConfig['ws.acl.enforce.endpoint.patterns'] || [];
-
 interface WorkspaceAuthorizer {
   authorizeWorkspace: (
     request: OpenSearchDashboardsRequest,
     workspaceIds: string[],
     permissionModes?: string[]
   ) => Promise<{ authorized: true } | { authorized: false; unauthorizedWorkspaces: string[] }>;
+  aclEnforceEndpointPatterns: string[];
 }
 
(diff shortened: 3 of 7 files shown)


server/services/AlertService.js
```diff
@@ -14,7 +14,17 @@ export const GET_ALERTS_SORT_FILTERS = {
 };
 
 export default class AlertService extends MDSEnabledClientService {
+  _enforceWorkspaceAcl = async (context, req, res, permissionModes) => {
+    const authorized = await this.checkWorkspaceAcl(context, req, permissionModes);
+    if (!authorized) {
+      return res.ok({ body: { ok: false, resp: 'Workspace ACL check failed: unauthorized' } });
```
huyaboo: Let's not mark the response as `ok` and instead use `res.unauthorized`

server/services/AlertService.js
```diff
@@ -14,7 +14,17 @@ export const GET_ALERTS_SORT_FILTERS = {
...
+    }
+    return null;
```
huyaboo: Lets use `undefined` as defined in the interface

server/services/MDSEnabledClientService.ts
```diff
@@ -1,12 +1,63 @@
-import { RequestHandlerContext, OpenSearchDashboardsRequest, ILegacyCustomClusterClient } from '../../../../src/core/server';
+import { RequestHandlerContext, OpenSearchDashboardsRequest, ILegacyCustomClusterClient, Logger } from '../../../../src/core/server';
+
+interface WorkspaceAuthorizer {
+  authorizeWorkspace: (
+    request: OpenSearchDashboardsRequest,
+    workspaceIds: string[],
+    principal: string,
+    permissionModes?: string[]
+  ) => Promise<{ authorized: boolean; unauthorizedWorkspaces?: string[] }>;
```
huyaboo: Lets use the `WorkspaceAuthResult` defined in the interface

server/services/MDSEnabledClientService.ts
```diff
@@ -1,12 +1,63 @@
...
+    this.workspaceStart = workspaceStart;
+  }
+
+  public setWorkspaceIdGetter(fn: (request: OpenSearchDashboardsRequest) => string | undefined) {
+    this.workspaceIdGetter = fn;
+  }
+
+  public setLogger(logger: Logger) {
+    this.logger = logger;
+  }
```
huyaboo, line 28: These should be set once at class instantiation time

server/services/MDSEnabledClientService.ts
```diff
@@ -1,12 +1,63 @@
...
+    if (dataSourceId) {
+      const savedObjectsClient = context.core.savedObjects.client;
+      const dataSource = await savedObjectsClient.get('data-source', dataSourceId.toString());
+      const endpoint = (dataSource.attributes as any).endpoint as string;
+      if (!endpoint.includes('.aoss.amazonaws.com')) {
+        return true;
+      }
+    } else {
+      return true;
+    }
```
huyaboo: Similar comments to the other PR: https://github.com/opensearch-project/dashboards-notifications/pull/445

server/services/MonitorService.js
```diff
@@ -11,8 +11,23 @@ import { MDSEnabledClientService } from './MDSEnabledClientService';
...
+   * Checks workspace ACL and returns an unauthorized response if check fails.
+   * Returns null if authorized, or a response object if unauthorized.
+   */
+  _enforceWorkspaceAcl = async (context, req, res, permissionModes) => {
+    const authorized = await this.checkWorkspaceAcl(context, req, permissionModes);
+    if (!authorized) {
+      return res.ok({ body: { ok: false, resp: 'Workspace ACL check failed: unauthorized' } });
+    }
+    return null;
+  };
```
huyaboo: If both child classes implement the same method, consider just putting it as part of the parent

server/services/MDSEnabledClientService.ts
nishtham-amazon: The services (MonitorService, AlertService, etc.) are instantiated during setup(), but the workspace plugin's start contract (authorizeWorkspace) isn't available until start(). OSD plugin lifecycle is: all      
  plugins setup() first, then all plugins start() — so plugins.workspace doesn't exist when these services are constructed.                                                                                          
                                                                                                                                                                                                                     
  These setters are only called once from start() and never again. Happy to consolidate them into a single initialize({ workspaceStart, workspaceIdGetter, logger }) method if that makes the one-time intent        
  clearer, but the timing can't move to construction. 

Let me know if you see this as a concern.

server/services/MonitorService.js
```diff
@@ -35,6 +37,8 @@ export default class MonitorService extends MDSEnabledClientService {
 
   createWorkflow = async (context, req, res) => {
```
eirsep, line 38: workflow APIs are not supported

eirsep: in fact, we would need some kind of check to block the workflow APIs in dashboards.

server/services/MonitorService.js
```diff
@@ -534,6 +561,8 @@ export default class MonitorService extends MDSEnabledClientService {
 
   acknowledgeChainedAlerts = async (context, req, res) => {
```
eirsep, line 562: not supported

server/alerting_configs.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ws.acl.enforce.endpoint.patterns": [".aoss.amazonaws.com"]
```
eirsep: 1. @huyaboo is this ok to be exposed? is AD doing the same? 
2. why is this `ws.acl.enforce.endpoint.patterns` plugin specific. should this come from dashboards core ?

eirsep (changes requested):

server/alerting_configs.json
huyaboo: 1. No, we ideally should include vendor-specific configurations as part of the `opensearch_dashboards.yml`. See this comment: https://github.com/opensearch-project/dashboards-notifications/pull/445/changes/BASE..9629a0ec33b073908614f990d9bfe8c671fb04de#diff-6d8c874001196bd53867951ab30dad3dcf29dbe5f7183a5fd566b5ac8ad37975
2. ~~So it seems like this config is scoped to alerting only, and it should be a alerting specific configuration~~ Actually reading again, this config should be added as part of a workspace config set in OSD core

riysaxen-amzn (approved):

server/services/MonitorService.js
AWSHurneyt: @eirsep The workflow UI will be hidden in the UI when the data source is a collection. The only other places in the UI that would call the workflows API would be devtools, but the backend would simply fail to return a response when such an API is called.

AWSHurneyt (approved):

eirsep (approved):

PR #1460, opened 2026-06-02
PPL Monitor Creation in Logs

TackAdam:
# Defer "Create monitor" Explore action when observability owns it

## Description

Lets the alerting plugin step aside on Explore's "Actions" menu when `dashboards-observability` has claimed ownership of the Create monitor entry, so users don't see two duplicate entries when both plugins are loaded.

## What changed

- Added `observabilityDashboards` to `optionalPlugins` in `opensearch_dashboards.json`.
- Extended `AlertingSetupDeps` with an optional `observabilityDashboards?: { ownsMonitorCreation: boolean }`. The shape is inlined (not imported from the observability package) so this PR doesn't introduce a hard cross-plugin type dependency.
- Extracted the registration predicate into `public/utils/should_register_alerting_explore_action.ts` and replaced the inline gate in `plugin.tsx` with a call to it. Same behavior, but the predicate is now testable in isolation without spinning up the full plugin lifecycle.
- The existing `explore.queryPanelActionsRegistry.register({...})` call in `setup()` is now guarded by `shouldRegisterAlertingExploreAction({ explore, observabilityDashboards })`.

## Why

The companion change in `dashboards-observability` registers its own Create monitor action on Explore that opens the observability alert-manager flyout (PPL-only, with inline datasource/index/time-field pickers and a "Run preview" affordance). When `observability.alertManager.enabled: true`, observability's flag is the source of truth and we want only its entry visible. When the flag is off — or observability isn't installed — alerting's existing entry behaves exactly as before.

## Behavior

| Observability loaded? | `observability.alertManager.enabled`? | Result |
| --- | --- | --- |
| No | n/a | Alerting's entry only (status quo) |
| Yes | `false` | Alerting's entry only — observability ships quiet |
| Yes | `true` | Alerting defers; observability's entry only |

## Tests

`public/utils/should_register_alerting_explore_action.test.ts` covers the predicate end-to-end. Six cases mirror the behavior matrix plus two defensive paths:

- does not register when Explore is absent
- does not register when Explore is absent even if observability is present
- registers when Explore is present and observability is absent
- registers when observability is loaded but did not claim monitor creation (the `alertManager.enabled: false` case)
- defers when observability has claimed monitor creation
- treats a missing `ownsMonitorCreation` field as "not claimed" — covers older observability builds that predate the contract

The test file uses a triple-slash `/// <reference types="jest" />` directive at the top so editors that don't already pick up `@types/jest` from the workspace resolve `describe` / `it`. No `tsconfig.json` change required (this plugin doesn't ship one).

## Backwards compatibility

The new dep is optional and the gate is structural. In any cluster where the companion observability change hasn't landed, `observabilityDashboards?.ownsMonitorCreation` is `undefined` → falsy → alerting registers as before. No-op on existing installations.

 
### Check List
  - [ ] All tests pass

### DIFF
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -8,6 +8,7 @@
     "dataSourceManagement",
     "assistantDashboards",
     "explore",
+    "observabilityDashboards",
     "workspace"
   ],
   "requiredPlugins": [
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -39,6 +39,10 @@ import type { ExplorePluginSetup, ExplorePluginStart } from '../../../src/plugin
 import { ResultStatus } from '../../../src/plugins/data/public';
 import { CreateMonitorFlyout } from './components/CreateMonitorFlyout';
 import { CoreContext } from './utils/CoreContext';
+import {
+  ObservabilityDashboardsSetupShape,
+  shouldRegisterAlertingExploreAction,
+} from './utils/should_register_alerting_explore_action';
 
 declare module '../../../src/plugins/ui_actions/public' {
   export interface ActionContextMapping {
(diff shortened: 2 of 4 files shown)


lezzago (dismissed):
Ideally add a unit test if possible

lezzago (approved):

riysaxen-amzn (approved):

PR #1461, opened 2026-06-05
Dynamic flag action

TackAdam:
### Description
The "Create monitor" entry on Explore's Query Panel "Actions" menu is registered by both this plugin and `dashboards-observability`. They previously coordinated via observability's setup-time `ownsMonitorCreation` boolean — read once at boot — so exactly one of them registered.

That boolean is no longer expressive enough. Observability is moving its "alerting on/off" decision from a static YAML flag to a request-time capability (`capabilities.observability.alertManagerEnabled`) that's seeded from YAML and overridable per request via `DynamicConfigService`. Operators flipping the YAML flag — or AppConfig overrides flipping it dynamically — should swap which plugin owns the menu entry without an OSD restart.

This PR moves our coordination off the setup-time boolean and onto the capability:

- We **always** load against Explore at setup time.
- We defer our `register(...)` call until `getStartServices()` resolves so capabilities are populated.
- We read `capabilities.observability.alertManagerEnabled`. If `true`, observability is showing its entry — we skip registration. Otherwise we register normally.

Observability does the mirror check on its side, so exactly one entry exists at any time. With both PRs deployed:

- YAML flips on/off → entry swaps after an OSD restart.
- AppConfig dynamic flag flips on/off → entry swaps on the next page load.
- Observability not installed at all → capability is absent → we register, exactly as before.

## Why defer to `getStartServices()`

Explore's `queryPanelActionsRegistry` exposes only `getIsEnabled` — which **disables** (greys out) an entry — and no `isVisible` hook. So if both plugins call `register(...)`, the menu shows two entries and one is greyed out. The only way to truly hide is to skip `register(...)` entirely. That requires reading the capability, which lives on `CoreStart`, not `CoreSetup`. Deferring registration to the resolved start services is the smallest change that gets us there. The promise resolves before any user can open the menu, so registration order isn't user-visible.

### `public/plugin.tsx`

- Removed the setup-time `if (shouldRegisterAlertingExploreAction({ explore, observabilityDashboards }))` gate.
- Replaced it with:

```ts
if (explore) {
  this.startServicesPromise.then(([coreStart]) => {
    // check capability, then register
  });
}
```

- The deferred callback reads `coreStart.application.capabilities.observability?.alertManagerEnabled`. Returns early when `true` (observability owns the entry). Otherwise registers our action with the existing `getIsEnabled` checks for `isPplAlertingEnabled()`, query result status, and AOSS collection guarding intact.
- Removed the now-unused `observabilityDashboards` destructuring from `setup()`'s deps and the `shouldRegisterAlertingExploreAction` import.
- Updated `AlertingSetupDeps.observabilityDashboards` typing to `unknown` with a comment pointing at the new capability-based coordination. The optional dependency declaration is preserved so OSD's plugin loader still resolves it when observability is installed.

### `public/utils/should_register_alerting_explore_action.ts` (deleted)

The helper's whole purpose was the setup-time gate we just removed. No other file referenced it.

### `public/utils/should_register_alerting_explore_action.test.ts` (deleted)

Tested the deleted helper.

## Behavior matrix

| Observability installed? | `capabilities.observability.alertManagerEnabled` | Result |
|---|---|---|
| No | Capability absent | Alerting registers. Only "Create monitor" shown. |
| Yes | `true` | Alerting skips. Observability shows "Create alert rule". |
| Yes | `false` | Alerting registers. Only "Create monitor" shown. |
| Yes | Not yet resolved | Alerting skips registration in this rare race window — won't lead to duplicates because observability hasn't registered yet either. |

Capability is YAML-seeded by `dashboards-observability` (`observability.alertManager.enabled: <bool>` in `opensearch_dashboards.yml`) and overridable per request via `DynamicConfigService`. We don't read YAML directly here — we trust the capability the observability plugin publishes.

## Backwards compatibility

`AlertingSetupDeps` still declares `observabilityDashboards?` so older observability builds still resolve as a registered optional dependency. We just don't read fields off it anymore. Older observability builds that still set `ownsMonitorCreation` will continue working with older releases that consumed it — this PR only changes how *this* plugin coordinates going forward.

The user-visible result for any existing deployment is the same as before — the only change is that the menu entry now responds to capability changes instead of being frozen at boot.

## Files changed

- `public/plugin.tsx` — defer `register()` call, read capability, decide.
- `public/utils/should_register_alerting_explore_action.ts` — deleted.
- `public/utils/should_register_alerting_explore_action.test.ts` — deleted.

### Local

1. Install both this plugin and `dashboards-observability` against the same OSD checkout.
2. Set `observability.alertManager.enabled: true` in `opensearch_dashboards.yml`. Restart. Open Explore → Actions menu. Should see only "Create alert rule" (observability's).
3. Flip to `observability.alertManager.enabled: false`. Restart. Should see only "Create monitor" (this plugin's).
4. Uninstall observability entirely. Should see only "Create monitor" (no capability published, we register).

### With dynamic config (downstream / AppConfig deployment)

1. YAML has `observability.alertManager.enabled: false`, but the AppConfig profile sets it to `true` for the test account.
2. Reload the page. Should see "Create alert rule" — capability resolves to `true`, observability registers, we skip.
3. Flip the AppConfig value back to `false`. Reload. Should see "Create monitor".
 
### Check List
  - [ ] All tests pass


riysaxen-amzn:
### 1. Race condition: deferred registration timing (Medium)

The PR assumes `startServicesPromise` resolves before a user can open the Explore Actions menu. That's probably true in practice, but it's an implicit ordering guarantee that isn't enforced.

If `getStartServices()` is slow for any reason (a heavy plugin blocking start, network latency on a capability switcher call), a user navigating directly to Explore could open the Actions menu and see *no* "Create monitor" entry at all — neither plugin has registered yet.

**Suggestion:** Add a brief comment acknowledging this window, or consider registering optimistically. If the registry doesn't support unregistration, the current approach is the best available — just document the assumption.

### 2. Type assertion on capabilities (Low)

```typescript
const capabilities = coreStart.application.capabilities as
  | { observability?: { alertManagerEnabled?: boolean } }
  | undefined;
```

`coreStart.application.capabilities` is never `undefined` — it's always a `Capabilities` object. The `| undefined` branch is dead. Consider dropping it.

### 3. No test coverage for the new deferred path (Medium-High)

The deleted test file covered the old helper. The new deferred-registration-with-capability-check logic has no new test coverage. Key scenarios to pin:

- `alertManagerEnabled === true` → `register()` is NOT called.
- `alertManagerEnabled === false` → `register()` IS called.
- `alertManagerEnabled` absent (observability not installed) → `register()` IS called.
- `startServicesPromise` rejects → what happens? Worth a test.

The PR checklist says "New functionality includes testing" but the net is −150 lines of tests.

### 4. Strict equality check — add a comment (Low)

```typescript
if (capabilities?.observability?.alertManagerEnabled === true) {
  return;
}
```

This strict `=== true` check is correct since capabilities values can be non-boolean. Worth a one-line comment to prevent a future contributor from "simplifying" it to a truthy check.

### 5. Non-null assertion inside callback (Nit)

```typescript
explore!.queryPanelActionsRegistry.register({
```

Inside the `.then()` callback, TypeScript can't narrow the outer `explore`. Consider capturing it before the closure:

```typescript
if (explore) {
  const explorePlugin = explore;
  this.startServicesPromise.then(([coreStart]) => {
    explorePlugin.queryPanelActionsRegistry.register({
```

Avoids the `!` entirely.

---

**Overall:** Main ask is adding unit tests for the new capability-gated registration path. Everything else is nit-level.

TackAdam:
Thanks for the thorough review  @riysaxen-amzn — addressed all five.

### 1. Race condition (Medium)

Added a paragraph in the deferred-registration comment block acknowledging the window: in the common case `getStartServices()` resolves before any user can open the Explore Actions menu, but on slow-start hosts a user navigating straight to Explore could briefly see no entry until the promise settles.

The registry does not support unregistration, so registering optimistically and tearing down later is not an option. Deferring registration remains the correct trade-off given the current constraints.

### 2. Dead `| undefined` branch (Low)

Dropped. `coreStart.application.capabilities` is always defined.

### 3. Test coverage (Medium-High)

Added `public/plugin.test.ts` with five tests covering:

- `alertManagerEnabled === true` → `register()` is **not** called.
- `alertManagerEnabled === false` → `register()` **is** called.
- Observability capability absent (plugin not installed) → `register()` **is** called.
- `startServicesPromise` rejects → no throw, no registration, error logged via the new `.catch()`.
- Explore plugin absent → `register()` is not called.

### 4. Strict `=== true` comment (Low)

Added a one-line comment beside the check warning future contributors not to "simplify" it to a truthy check, since capability values may not always be strictly boolean.

### 5. Non-null assertion (Nit)

Captured `explore` as `explorePlugin` before entering the `.then()` closure so TypeScript narrows correctly within the callback. The non-null assertion (`!`) is no longer needed.

### Bonus

Added a `.catch()` at the end of the deferred registration chain so a `getStartServices()` rejection surfaces as a `console.error` instead of becoming an unhandled promise rejection.

This follows the same pattern used in the companion observability PR. Test #4 above validates the behavior.

lezzago:
A few minor observations — overall the design is correct and the two-PR sequence is safe to deploy in any order.

---

**1. Indentation bug in `plugin.tsx` — `.then()` callback body misaligned**

The body of the `.then()` callback sits at the same indent level as the `.then(` call itself rather than one level deeper. This will likely fail the project's ESLint/Prettier check and makes the `.catch(` visually appear to chain off the outer `if (explore)` block instead of the promise.

---

**2. `flushMicrotasks` in tests is fragile (`public/plugin.test.ts`)**

```ts
async function flushMicrotasks() {
  await Promise.resolve();
  await Promise.resolve();
}
```

Two hardcoded ticks work for today's call depth, but if the registration path ever gains one more `await`, all tests will silently pass without executing the callback. `jest.runAllTicks()` or `await new Promise(process.nextTick)` would be more robust. At minimum, a comment explaining why exactly two ticks are sufficient would help future readers.

---

**3. `observabilityDashboards?: unknown` loses documentation value**

Typing it as `unknown` is correct (we don't read it), but discards context about what older builds put here. Consider:

```ts
observabilityDashboards?: { ownsMonitorCreation?: boolean; [key: string]: unknown };
```

This preserves the backward-compat story for future developers without importing from the deleted module.

---

**Deployment note:** If this PR deploys before the observability PR, `capabilities?.observability?.alertManagerEnabled` resolves to `undefined` and the strict `=== true` check correctly falls through to register alerting's action — no duplicates, no missing entries. Good defensive choice.

lezzago (approved):
Design is correct, backward-compat is solid, and deployment order is safe. Minor nits in the comment above — none are blocking.

riysaxen-amzn (approved):

lezzago:
**Indentation bug in `public/plugin.tsx` — `.then()` callback body misaligned**

The body of the `.then()` callback is indented at the same level as the `.then(` call itself rather than one level deeper:

```ts
this.startServicesPromise
  .then(([coreStart]) => {
  const capabilities = ...   // ← should be indented one more level
  if (...) { return; }
  explorePlugin.queryPanelActionsRegistry.register({
    ...
  });
})
  .catch(...)
```

This will likely fail ESLint/Prettier. It also makes the `.catch(` visually appear to chain off the outer `if (explore)` block instead of the promise.

PR #1463, opened 2026-06-09
Drop observability dependency 

TackAdam:
### Description

Drops the now-dead `observabilityDashboards` optional plugin dependency.

The original setup-time coordination between this plugin and `dashboards-observability` for the Explore "Create monitor" / "Create alert rule" menu entry was unwound in #1461. That PR moved coordination to a request-time capability check (`capabilities.observability.alertManagerEnabled`) and stopped reading anything from `setupDeps.observabilityDashboards`.

The companion observability PR ([opensearch-project/dashboards-observability#2719](https://github.com/opensearch-project/dashboards-observability/pull/2719)) completes the cleanup on its side by removing `ownsMonitorCreation` from its setup contract entirely.

Once that lands, this plugin has zero consumers of the observability setup contract, making the optional dependency effectively a phantom dependency.

### What's preserved

The runtime "which button should be shown" behavior remains unchanged.

Ownership continues to be determined through the capability published by observability's `registerSwitcher` and consumed from:

```ts
coreStart.application.capabilities.observability.alertManagerEnabled
```

inside the deferred:

```ts
getStartServices().then(...)
```

registration path.

This mechanism operates entirely through `CoreStart` capabilities and no longer requires a plugin-level dependency.

### `public/plugin.tsx`

- Removed `observabilityDashboards?: unknown` from `AlertingSetupDeps`.
- The setup dependency interface now reflects reality: alerting no longer consumes any setup contract from observability.

### `opensearch_dashboards.json`

- Removed `"observabilityDashboards"` from `optionalPlugins`.
- The plugin manifest now reflects the actual runtime dependency graph.

## Risk

None.

The field was already typed as `unknown` and never read. The manifest entry was only load-order metadata for a relationship that no longer exists.

## Testing

- `yarn test:jest public/plugin.test public/plugin_nav`

Results:

- All 7 existing tests pass.
  - 5 deferred-registration coordination tests
  - 2 navigation tests

## Issues Resolved

N/A — cleanup follow-up to #1461.
 
### Check List
  - [ ] All tests pass


lezzago:
**Indentation bug in `public/plugin.tsx` — `.then()` callback body misaligned**

The body of the `.then()` callback is indented at the same level as the `.then(` call itself rather than one level deeper:

```ts
this.startServicesPromise
  .then(([coreStart]) => {
  const capabilities = ...   // ← should be indented one more level
  if (...) { return; }
  explorePlugin.queryPanelActionsRegistry.register({
    ...
  });
})
  .catch(...)
```

This will likely fail ESLint/Prettier. It also makes the `.catch(` visually appear to chain off the outer `if (explore)` block instead of the promise. Since #1461 and #1463 share the same branch, fixing it once resolves it for both.

TackAdam:
Closing in favor of #1464, which is opened from the correct branch (#1463 was inadvertently still tracking the merged branch from #1461).

PR #1496, opened 2026-08-08
Integrate centralized resource-sharing share button for monitors and workflows

DarshitChanpura:
### Description
Integrates the centralized share button from [security-dashboards-plugin PR #2491](https://github.com/opensearch-project/security-dashboards-plugin/pull/2491) via its dependency-free DOM-marker SPI, surfacing per-monitor sharing directly in the monitors list.

**How it works**
- A Share column in the monitors table renders marker elements (`data-resource-share-button` + id/type attributes); the security dashboards plugin discovers them and mounts its centralized Share/Update Access button + modal
- Rows resolve their resource type per item: composite monitors map to the `workflow` type, all other monitors to `monitor` — matching the types registered by `AlertingResourceSharingExtension` in [alerting#2180](https://github.com/opensearch-project/alerting/pull/2180)
- The column is gated on the core `resourceSharing` capability (`enabled` + type present in `availableTypes`), so it is completely absent when the security dashboards plugin is not installed, the feature is disabled, or the alerting backend does not yet contain #2180 — no dependency on the security plugin (no manifest changes, no imports)

### Draft status
Blocked on two upstream merges: security-dashboards-plugin#2491 (provides the SPI + capability) and alerting#2180 (registers the types). Safe to merge before either — the column stays dormant until both are present.

### Testing
- Compiled and loaded against OSD main + OpenSearch 3.8.0 with resource sharing enabled; column correctly absent while the backend lacks #2180
- Share flow semantics (modal, can_share graying, feature-off hiding) covered by the security-dashboards-plugin PR's unit/integration tests

### Check List

### DIFF
--- a/public/pages/Monitors/containers/Monitors/Monitors.js
+++ b/public/pages/Monitors/containers/Monitors/Monitors.js
@@ -136,7 +136,7 @@ export default class Monitors extends Component {
               // button is mounted here by security-dashboards-plugin when
               // installed and resource sharing is enabled for monitors.
               field: 'id',
-              name: 'Share',
+              name: 'Access',
               sortable: false,
               width: '50px',
               render: (id, item) => {
@@ -149,6 +149,7 @@ export default class Monitors extends Component {
                     data-resource-share-button
                     data-resource-id={id}
                     data-resource-type={resourceType}
+                    {...(item.name ? { 'data-resource-name': item.name } : {})}
                     data-resource-share-display="icon"
                   />
                 ) : null;
--- a/public/pages/Monitors/containers/Monitors/Monitors.js
+++ b/public/pages/Monitors/containers/Monitors/Monitors.js
@@ -25,7 +25,12 @@ import {
   isDataSourceChanged,
   getDataSourceId,
 } from '../../../utils/helpers';
-import { getUseUpdatedUx } from '../../../../services';
+import {
+  ALERTING_WORKFLOW_RESOURCE_TYPE,
+  getUseUpdatedUx,
+  isResourceSharingAvailable,
+  MONITOR_RESOURCE_TYPE,
+} from '../../../../services';
 
 const MAX_MONITOR_COUNT = 1000;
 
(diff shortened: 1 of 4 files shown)


DarshitChanpura:
Screenshot 2026-08-26 at 2 21 16 PM

riysaxen-amzn:
Nice clean approach — the dependency-free DOM-marker SPI with capability gating keeps this plugin decoupled from security-dashboards-plugin, and the availability helper + column tests are appreciated. A few items, one blocking:

## 1. Blocking: workflow resource type no longer matches alerting#2180

`public/services/services.ts`:

```ts
export const ALERTING_WORKFLOW_RESOURCE_TYPE = 'workflow';
```

alerting#2180 renamed the registered workflow type to **`alerting-workflow`** (commit 5798725f, "Rename workflow RSC resource type to alerting-workflow to avoid flow-framework collision") — see `ResourceSharingUtils.WORKFLOW_RESOURCE_TYPE` on that branch. As written here, composite monitor rows emit `data-resource-type="workflow"`, which in the shared registry is **flow-framework's** type:

- `isResourceSharingAvailable('workflow')` checks the wrong type's availability
- the mounted share modal would resolve sharing entries against the wrong resource registry

The constant is even named `ALERTING_WORKFLOW_...`, so I suspect the value just predates the rename. The unit test in `Monitors.test.js` asserts `'workflow'` too, so it cements the drift rather than catching it. Fix: value → `'alerting-workflow'` in the constant, the test expectation, and the `availableTypes` fixtures.

## 2. Question: composite detection via `item.monitor?.type === 'workflow'`

Can you confirm this matches how the monitors list actually represents workflows? Other parts of this codebase distinguish via `monitor_type` / `workflow_type` fields. If composite list items don't carry `monitor.type === 'workflow'`, they'd silently fall through to the `monitor` type and share against the wrong registry entry.

## 3. Question: MDS (multi-data-source) interaction

The monitors listed can come from a **remote** data source, but the `resourceSharing` capability and the mounted share modal come from the **local** cluster's security plugin. Sharing a remote monitor's ID against the local sharing registry seems wrong (or at best a no-op against a nonexistent resource). Should the column be gated on the local cluster being the active data source (e.g. `!getDataSourceId()`), or does the centralized button handle `dataSourceId`?

## 4. Minor: column gate asymmetry

The Access column only renders when `isResourceSharingAvailable(MONITOR_RESOURCE_TYPE)`. If only the workflow type were in `protected_types`, workflow rows would get no share column at all. Cheap fix: gate the column on monitor **or** workflow availability — the per-row check already handles the rest.

## Nits

- Column header says "Access" while the description calls it a Share column — pick one for consistency.
- `width: '120px'` is generous for `data-resource-share-display="icon"`.

(CI note: the unit-test failures are unrelated infra — the job dies with `bash: yarn: command not found` after the nvm Node 22.23.2 install, before any tests run.)

DarshitChanpura:
Thanks for the review suggestions. Addressed in `badf579`:

1. **Include the workflow type in the Access column availability check** — done. The outer guard now renders the column when either `MONITOR_RESOURCE_TYPE` **or** `ALERTING_WORKFLOW_RESOURCE_TYPE` is shareable, so composite (workflow) monitor rows keep their share button even when only the workflow type is enabled. Per-row gating still selects the correct resource type for each row. Added a `renders the Access column when only the workflow type is available` test.

2. **Fragile `availableTypes` parsing** — added `.map(t => t.trim())` so values like `"monitor, workflow"` resolve correctly, with a whitespace-tolerant test. I left out the array-handling branch: `availableTypes` is produced by the security plugin's SPI capability as a comma-separated string, so the array form can't occur by design and defending against it would add dead code.

DarshitChanpura:
CI will be fixed by #1507

riysaxen-amzn (approved):

PR #1505, opened 2026-08-26
Accept v1-shape monitor body in PPL update guard and toV1MonitorBody

riysaxen-amzn:
The monitor details page enable/disable toggle round-trips the monitor in raw v1 shape (query nested in inputs[0].ppl_input.query, nothing at top level). toV1MonitorBody read only the top-level query and defaulted to '', so before the empty-query guard this path silently wiped the monitor's PPL query on every toggle; after the guard it was rejected outright.

Fall back to inputs[0].ppl_input.query in both toV1MonitorBody and the updateMonitor guard so v1-shape bodies pass through with their query intact, while truly query-less bodies are still rejected.

### Description
[Describe what this change achieves]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/server/services/PplAlertingMonitorService.js
+++ b/server/services/PplAlertingMonitorService.js
@@ -15,13 +15,13 @@ const ALERTS_BASE_PATH = `${PPL_MONITOR_BASE_API}/alerts`;
 
 /**
  * Transforms the frontend body format ({ ppl_monitor: { name, query, triggers, ... } })
- * into the v1 backend format ({ name, monitor_type, inputs, triggers, ... }).
+ * into the engine's native monitor format ({ name, monitor_type, inputs, triggers, ... }).
  */
-const toV1MonitorBody = (body) => {
+const toEngineMonitorBody = (body) => {
   const pplMon = body?.ppl_monitor || body;
   // Accept both the flattened frontend shape (query at top level) and the
-  // raw v1 shape (query nested in inputs[0].ppl_input.query). Callers that
-  // round-trip a v1-format monitor (e.g. the monitor details page
+  // raw engine shape (query nested in inputs[0].ppl_input.query). Callers that
+  // round-trip an engine-format monitor (e.g. the monitor details page
   // enable/disable toggle) send the latter.
   const query = pplMon.query || pplMon.inputs?.[0]?.ppl_input?.query || '';
   const rawTriggers = Array.isArray(pplMon.triggers) ? pplMon.triggers : [];
@@ -30,7 +30,7 @@ const toV1MonitorBody = (body) => {
     return { ppl_trigger: t };
   });
 
-  const v1 = {
+  const engineMonitor = {
     name: pplMon.name,
     monitor_type: 'ppl_monitor',
     enabled: pplMon.enabled !== false,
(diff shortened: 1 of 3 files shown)


server/services/PplAlertingMonitorService.test.js
```diff
@@ -102,6 +102,61 @@ describe('PplAlertingMonitorService.updateMonitor query guard', () => {
     expect(callArgs.body.inputs[0].ppl_input.query).toContain('source = logs-*');
     expect(callArgs.body.triggers[0].ppl_trigger).toBeDefined();
   });
+
+  test('accepts a raw v1-shape body whose query is nested in inputs[0].ppl_input (details-page enable/disable round-trip)', async () => {
```
eirsep: there's no test sending inputs: [{ ppl_input: {} }] (or an empty nested  query) and asserting rejection?

eirsep (approved):

server/services/PplAlertingMonitorService.js
```diff
@@ -616,8 +620,12 @@ export default class PplAlertingMonitorService extends MDSEnabledClientService {
 
       // Guard against silent data loss: toV1MonitorBody defaults a missing
       // query to '', so an update body without a query would silently wipe
-      // the monitor's PPL query. Reject it instead.
-      if (!cleanMonitor.query || !String(cleanMonitor.query).trim()) {
+      // the monitor's PPL query. Accept the query from either the flattened
+      // shape (top-level query) or the raw v1 shape (inputs[0].ppl_input.query,
```
toepkerd: nit: let's avoid using V1 or V2 language as this is a vestige of a time where we were pushing for PPL Monitors to be a wave of new V2 APIs.

toepkerd (approved):

toepkerd:
Some of the unit test CIs appear to fail due to compiler errors. Are these outdated CIs?

server/services/PplAlertingMonitorService.test.js
riysaxen-amzn: Good catch -- added two tests in 1d963a8: one with `inputs: [{ ppl_input: { query: '' } }]` and one with `inputs: [{ ppl_input: {} }]`, both asserting the guard rejects with the same message and the engine client is never called.

server/services/PplAlertingMonitorService.js
riysaxen-amzn: Agreed - 1d963a8 removes all v1/v2 language from the files this PR touches: toV1MonitorBody -> toEngineMonitorBody, flattenV1Monitor -> flattenEngineMonitor, and all comments reworded to "engine format" vs "flattened format". Broader cleanup of the remaining v2 vestiges (the /api/alerting/v2 route namespace.

lezzago:
## Code review findings

Reviewed the nested-query fallback fix and the `toEngineMonitorBody`/`flattenEngineMonitor` rename. The core change is correct — the guard and translator both read from the same source (`query` → `inputs[0].ppl_input.query`), so they stay consistent, and all 8 references to the renamed functions are updated with no stale references. Three items surfaced:

### 1. Stale trigger metadata survives on the newly-accepted engine shape (medium)
`server/services/PplAlertingMonitorService.js` — the trigger-cleaning map strips `id`/`last_triggered_time`/`last_execution_time` only from **unwrapped** triggers:

```js
cleanMonitor.triggers.map(({ id, last_triggered_time, last_execution_time, ...trigger }) => trigger)
```

The engine-shape body this PR now accepts wraps triggers as `[{ ppl_trigger: { id, last_triggered_time, ... } }]`. The destructure finds none of those keys at the wrapper's top level, so the inner `ppl_trigger` (with its stale `id` and last-run timestamps) passes through untouched — while the flattened path drops them. This is exercised by the new round-trip test (`trigger-id-1` is forwarded verbatim). Worth confirming the engine tolerates a stale trigger `id` on update, or unwrapping `ppl_trigger` before stripping so both shapes are cleaned consistently.

### 2. Duplicated query-extraction expression (cleanup)
`x.query || x.inputs?.[0]?.ppl_input?.query` is duplicated in the `updateMonitor` guard and in `toEngineMonitorBody`. They must stay identical or the guard could accept a body the translator then wipes to `''` (the exact silent-data-loss class this PR prevents). Extracting a single `extractPplQuery(pplMon)` helper used by both would prevent future drift.

### 3. Whitespace top-level query short-circuits the nested fallback (low)
In the guard, `cleanMonitor.query || nested` means a truthy-but-whitespace top-level `query` (e.g. `'   '`) short-circuits before the nested fallback, so the update is rejected even when a valid nested query is present. Very low real-world likelihood since the flatten path keeps top-level and nested in sync, but the `||` chaining makes the fallback unreachable whenever the top-level slot holds a blank string.

Overall: small, well-targeted PR. All changes are relevant (the rename is a reviewer-requested cleanup, fully self-contained). Only finding #1 is worth verifying before merge.

PR #1512, opened 2026-09-14
Guard the Access column against a data-source staleness race

DarshitChanpura:
### Description

`getResourceSharingAvailableTypes(landingDataSourceId)` is called asynchronously whenever the selected data source changes, but the resolved types were stored without recording which data source they belong to. On a data-source switch, the previous data source's resolved types could still decide the Access column's visibility in `Monitors.js` for the duration of the new, in-flight probe.

Found while addressing [kaituo's review](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/1238) on anomaly-detection-dashboards-plugin#1238, and confirmed to be missing from every resource-sharing consumer plugin on inspection — not specific to this plugin.

### Fix

`loadResourceSharingAvailability` now captures the requested data source id before awaiting `getResourceSharingAvailableTypes` and only applies the result via `setState` if it still matches `this.props.landingDataSourceId` when the promise resolves. Both render-time usages in `buildColumns()` — the column-visibility check (covering monitor OR workflow resource types) and the per-row render callback (checking the specific resource type for that row) — additionally guard against `resourceSharing.dataSourceId` no longer matching the currently selected data source.

### Category
Bug fix

### Testing
Updates the existing test helper to set `resourceSharing` as a `{dataSourceId, types}` tuple matching the component's `landingDataSourceId` prop, and adds a new test exercising the guard directly: a `resourceSharing` value resolved for one data source must not decide the Access column's visibility when a different data source is currently selected.

### Check List
- [x] New functionality includes testing
- [x] Commits are signed per the DCO using `--signoff`


DarshitChanpura:
### Re: Code-Diff-Analyzer flag on `opensearch_dashboards.json`

The analyzer flagged the new `"securityDashboards"` entry in `optionalPlugins` per its mandatory "flag all new dependency changes" rule — this is expected, not a defect.

**For maintainer review:**
- `securityDashboards` resolves to the in-repo `security-dashboards-plugin` (relative import, e.g. `../../security-dashboards-plugin/public/types`) — not an npm package, so there is no typosquatting/namespace-hijack surface.
- It is declared in `optionalPlugins`, not `requiredPlugins`, matching the existing pattern already used for `assistantDashboards`/`dataSource`/etc. in this manifest.
- The dependency is optional at every level: an optional TypeScript field (`securityDashboards?: SecurityPluginStart`), and a runtime `if (securityDashboards)` guard in `start()`. When security-dashboards-plugin is absent or disabled, that block never runs and the resource-sharing helper fails closed to `[]` without ever calling into it — covered by a dedicated unit test (`returns [] without probing when security-dashboards-plugin is not installed`).
- The same dependency + pattern is added identically across the sibling PRs below, so a single review here covers the shape of the change everywhere it appears.

Once reviewed, the standard unblock is: add the `skip-diff-analyzer` label, then re-run the failed check.

DarshitChanpura:
Closing this draft PR. The premise was incorrect: this plugin's per-data-source resource-sharing check was already correct, and gating it further on security-dashboards-plugin's local resource-sharing flag introduced a regression (the Access column would incorrectly disappear when local is disabled but the selected data source has sharing enabled).

The real defect (reported by kaituo on anomaly-detection-dashboards-plugin#1238) was in security-dashboards-plugin itself: its Share-button DOM-marker SPI watcher only started when the local cluster had resource sharing enabled. That is fixed directly in opensearch-project/security-dashboards-plugin#2525 (the SPI watcher now always starts), which requires no changes in this plugin.

DarshitChanpura:
Reopened with a different fix than what the earlier bot analysis/review comments above refer to (those were against the now-reverted local-SPI-gate approach on commit 87ed02f).

This PR now adds a data-source staleness guard to the Access column's resource-sharing availability check in `Monitors.js` — see the updated description above. No `securityDashboards` optional dependency or manifest changes are part of this PR.

riysaxen-amzn:
The staleness guard commit (`b55b0d7`) looks correct to me, but the branch does not match the description yet, so requesting changes until it is rebased.

**Blocking: `87ed02f` is still on the branch**

The Sep 15 comment says "No `securityDashboards` optional dependency or manifest changes are part of this PR", but the PR head still contains `87ed02f` "Gate the Access column on the local resource-sharing SPI". Files Changed shows 6 files, and these four are all from that commit:

- `opensearch_dashboards.json` (adds `securityDashboards` to `optionalPlugins`)
- `public/plugin.tsx` (`setSecurityDashboards` wiring)
- `public/services/services.ts` (local-SPI re-confirmation in `getResourceSharingAvailableTypes`, plus the `import { SecurityPluginStart } from '../../../security-dashboards-plugin/public/types'` relative import)
- `public/services/__tests__/resource_sharing.test.ts`

Please either drop `87ed02f` from the branch so the PR is only the staleness guard, or, if the SPI gate is meant to ship here, update the title and description and let it be reviewed on its own merits (it depends on opensearch-project/security-dashboards-plugin#2525 and the relative type import assumes the sibling plugin is checked out alongside).

**On the staleness guard itself (`Monitors.js`)**

- `loadResourceSharingAvailability` captures `requestedDataSourceId` before the await and only applies the result when it still equals `this.props.landingDataSourceId`. Correct.
- Both render-time uses in `buildColumns()` (column visibility and the per-row marker render) additionally check `resourceSharing.dataSourceId === landingDataSourceId`. Correct.
- The bot's "missing re-probe on data source switch" finding is not accurate: `componentDidUpdate` already calls `loadResourceSharingAvailability()` inside the `isDataSourceChanged(prevProps, this.props)` branch.

Nit: the new test seeds `resourceSharing` for `ds-a` via `setState` while `landingDataSourceId` is `ds-b`, which exercises the render guard but not the resolve-time guard. If you want to cover the actual race, mock `getResourceSharingAvailableTypes` with a deferred promise, change `landingDataSourceId`, then resolve it and assert `setState` was not applied. Not blocking.

riysaxen-amzn:
Approving. The rebase resolved the earlier concern: the PR is now a single commit (`0ceb2f9`) touching only `Monitors.js` and `Monitors.test.js`, so the branch matches the description and the `securityDashboards` optional dependency is no longer in scope here.

On the guard itself:

- `loadResourceSharingAvailability` captures `requestedDataSourceId` before the await and only applies the result when it still equals `this.props.landingDataSourceId`. Correct.
- Both render-time uses in `buildColumns()` (column visibility and the per-row marker render) additionally check `resourceSharing.dataSourceId === landingDataSourceId`. Correct.
- For anyone reading the earlier bot comment about a missing re-probe: `componentDidUpdate` already calls `loadResourceSharingAvailability()` inside the `isDataSourceChanged(prevProps, this.props)` branch, so a switch does trigger a fresh probe.

Nit, not blocking: the new test seeds `resourceSharing` for `ds-a` via `setState` while `landingDataSourceId` is `ds-b`, which covers the render guard but not the resolve-time guard. If you want the actual race covered, mock `getResourceSharingAvailableTypes` with a deferred promise, change `landingDataSourceId`, resolve it, and assert the state was not applied.

riysaxen-amzn (approved):

PR #1516, opened 2026-09-17
fix: make loopback window editable when editing a PPL monitor

vikhy-aws:
### Description
Makes the PPL monitor look back window work on the edit page. Previously, when editing a PPL monitor the look back window was not editable and the timestamp field showed blank (or only the previously selected value) instead of all available date fields. This change lets you edit the look back window on edit, shows all detected timestamp fields with the saved one selected, and defaults the timestamp field when re-enabling look back.
 
### Issues Resolved
N/A

### Screenshots
Screenshot 2026-09-17 at 6 44 33 AM
Screenshot 2026-09-17 at 6 44 44 AM
Screenshot 2026-09-17 at 6 45 03 AM

 
### Check List
- [ ] All tests pass


riysaxen-amzn:
**Blocking: switching the timestamp field on edit leaves the old time filter in the query**

Before this PR the `pplTimestampField` select was inside the `!isEdit` block, so it could never change on the edit page. Now it can, but `stripTimeFilterFromQuery` / `addTimeFilterToQuery` are keyed on the field name, and `pplAlertingMonitorToFormik` hydrates `pplQuery` straight from the persisted `monitor.query`, which already contains the injected `| where <savedField> > DATE_SUB(...)` clause.

I ran the helpers from `pplAlertingHelpers.js` against the persisted shape:

```
saved (created with @timestamp, 1h lookback):
  source=logs | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host

edit: change field to event_time, keep 1h, save:
  source=logs | where event_time > DATE_SUB(NOW(), INTERVAL 1 HOUR) | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host

edit: change field to event_time, disable lookback, save:
  source=logs | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host
```

In both cases the monitor keeps filtering on a field the user just deselected, and the result set is silently narrowed.

Two ways to fix:
1. In the `pplTimestampField` `onChange`, strip the clause for the previous field before setting the new one (`setFieldValue('pplQuery', stripTimeFilterFromQuery(pplQuery, previousField))`), or
2. Make `stripTimeFilterFromQuery` field-agnostic (match `| where <any identifier> > DATE_SUB(NOW(), INTERVAL ...)` and the legacy `TIMESTAMP(...)` form), since the plugin is the only thing that injects that exact shape.

Option 2 is simpler and also covers the case where the saved `timestamp_field` is missing or the field got renamed in the index. A unit test on `stripTimeFilterFromQuery` with a mismatched field name would lock it in.

riysaxen-amzn:
**Nit: `PplAlertingCreateMonitor` is not wired into any route**

```
$ grep -rn "PplAlertingCreateMonitor'" public server --include=*.js --include=*.ts --include=*.tsx | grep -v .test.
(no results)
```

Routing is `Main.js` → `CreateMonitor.js` → `DefineMonitor` + `PplScheduleEditor`, so the timestamp-preservation change here and `PplAlertingCreateMonitor.timestamp.test.js` exercise a container that never renders. Not a correctness problem, but it's worth either dropping these two files from the PR to keep the diff honest about what's live, or opening a follow-up to delete the container. Happy either way.

Related: the live container `CreateMonitor.renderPplSchedule` has a render-time reset at L208 that overwrites `timestampField` with `availableDateFields[0]` whenever the current value isn't in the detected list. That means the "include the saved field as a synthetic option" branch in `PplScheduleEditor` only holds while detection is in flight or errored; once fields load it gets reset anyway. Fine as a loading-state fallback, just flagging so the test name ("even when not detected") doesn't overstate the behaviour in the real page.

riysaxen-amzn:
**Nit: re-enable default and the `CreateMonitor` L208 reset overlap**

The new `if (!timestampField && availableDateFields.length > 0) setFieldValue('timestampField', availableDateFields[0])` on re-enable is correct, but `CreateMonitor.renderPplSchedule` L208 already does the same thing on every render when the value isn't in the list (and `''` never is). So today this fires twice with the same value. Harmless, but if the L208 reset is ever removed this becomes the only guard, so the test you added is the one that matters. Just noting so nobody deletes it as redundant later.

lezzago (commented):
Automated review of the PPL look-back editable-on-edit change. Four findings below. Two of them (the save-path double-filter and the render-time timestamp clobber) live in files this PR doesn't touch, so they're anchored here to the in-diff line that newly exposes them — the real location is cited in each comment.

public/components/PplScheduleEditor/PplScheduleEditor.js
```diff
@@ -148,7 +148,7 @@ const PplScheduleEditor = ({
       )}
 
       {/*// TODO: investigate whether we can add back the lookback window component for mustang domains*/}
-      {!isEdit && !isMustang && (
+      {!isMustang && (
```
lezzago, line 160: **Stale where-filter left behind when the timestamp field is changed on edit** (real location: `public/pages/CreateMonitor/containers/CreateMonitor/utils/pplFormikToMonitor.js:216`).

Making the look-back window and its timestamp-field select editable on edit newly exposes a save-path bug. On edit, `values.pplQuery` is loaded verbatim from `monitor.query` (`pplAlertingMonitorToFormik.js:40`) and already contains `| where old_ts > DATE_SUB(NOW(), INTERVAL ...)`. If the user switches the timestamp field (`old_ts` → `new_ts`) and saves, `addTimeFilterToQuery` → `stripTimeFilterFromQuery` (`pplAlertingHelpers.js:641`) builds its strip regex from **only the current** `timestampField`, so the `old_ts` clause survives and a `new_ts` clause is appended — the persisted query then double-filters on both fields.

Fix: when injecting/stripping the look-back filter on save, strip clauses for any prior timestamp field too (e.g. track the previous field, or strip all injected look-back `where` clauses regardless of field before re-adding).

public/components/PplScheduleEditor/PplScheduleEditor.js
```diff
@@ -37,7 +37,7 @@ const PplScheduleEditor = ({
   isMustang = false,
   wrapperStyle = { maxWidth: '720px' },
 }) => {
-  const useLB = isEdit ? false : useLookBackWindow !== undefined ? useLookBackWindow : true;
+  const useLB = useLookBackWindow !== undefined ? useLookBackWindow : true;
```
lezzago, line 39: **Dead `isEdit` prop.** With this line no longer branching on `isEdit` (and the `!isEdit` render guard at line 151 removed), `isEdit` now has no effect anywhere in this component — its only remaining reference is the destructure default at line 36. Yet `CreateMonitor.js:226` still passes `isEdit={this.props.edit}`, which misleads readers into thinking it controls behavior. Remove the prop from both the component and that call site.

public/components/PplScheduleEditor/PplScheduleEditor.js
```diff
@@ -148,7 +148,7 @@ const PplScheduleEditor = ({
       )}
 
       {/*// TODO: investigate whether we can add back the lookback window component for mustang domains*/}
-      {!isEdit && !isMustang && (
+      {!isMustang && (
         <>
```
lezzago, line 161: **Stale help text** (the `helpText` a few lines below, ~line 156). "After creating this monitor, that filter can be edited in the query itself." was written when the look-back window was create-only. This PR makes the window directly editable on the edit page and persists it as `look_back_window_minutes`, so the copy now contradicts the editable control rendered right next to it. Please update it to reflect that the window is editable here.

public/pages/CreateMonitor/containers/DefineMonitor/DefineMonitor.js
```diff
@@ -121,7 +121,13 @@ class DefineMonitor extends Component {
       if (hasTimeField || !this.requiresTimeField()) this.onRunQuery();
     }
     if (searchType === SEARCH_TYPE.CLUSTER_METRICS) this.getSupportedApiList();
-    if (searchType === SEARCH_TYPE.PPL) this.initializePplIndices();
+    if (searchType === SEARCH_TYPE.PPL) {
+      this.initializePplIndices();
+      // Without this the timestamp dropdown only shows the previously selected field instead of all available fields.
+      if (this.props.values?.pplQuery) {
+        this.detectPplTimestampFields(this.props.values.pplQuery);
```
lezzago, line 128: **This mount-time detection newly triggers a render-time side effect that clobbers the saved timestamp field** (real location: `public/pages/CreateMonitor/containers/CreateMonitor/CreateMonitor.js:208`).

Running `detectPplTimestampFields` on mount for edit flows through `notifyDateFieldsChange` → `handlePplDateFieldsChange` → populates `pplDateFields`. On the next render, `renderPplSchedule` (CreateMonitor.js:208) sees the saved field isn't in the freshly-detected `availableDateFields` (schema drift) and calls `setFieldValue('timestampField', availableDateFields[0])` **during render** — silently overwriting the monitor's saved timestamp field. This defeats the prepend fix added in this same PR (which was meant to keep showing the saved field when undetected) and mutates form state during render. It also feeds the double-filter bug above by changing the field on save.

Fix: guard line 208 so it only auto-selects when `timestampField` is unset (not merely undetected), and move the assignment out of render (e.g. into the detection callback / an effect).

vikhy-aws:
> **Blocking: switching the timestamp field on edit leaves the old time filter in the query**
> 
> Before this PR the `pplTimestampField` select was inside the `!isEdit` block, so it could never change on the edit page. Now it can, but `stripTimeFilterFromQuery` / `addTimeFilterToQuery` are keyed on the field name, and `pplAlertingMonitorToFormik` hydrates `pplQuery` straight from the persisted `monitor.query`, which already contains the injected `| where <savedField> > DATE_SUB(...)` clause.
> 
> I ran the helpers from `pplAlertingHelpers.js` against the persisted shape:
> 
> ```
> saved (created with @timestamp, 1h lookback):
>   source=logs | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host
> 
> edit: change field to event_time, keep 1h, save:
>   source=logs | where event_time > DATE_SUB(NOW(), INTERVAL 1 HOUR) | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host
> 
> edit: change field to event_time, disable lookback, save:
>   source=logs | where @timestamp > DATE_SUB(NOW(), INTERVAL 1 HOURS) | stats count() by host
> ```
> 
> In both cases the monitor keeps filtering on a field the user just deselected, and the result set is silently narrowed.
> 
> Two ways to fix:
> 
> 1. In the `pplTimestampField` `onChange`, strip the clause for the previous field before setting the new one (`setFieldValue('pplQuery', stripTimeFilterFromQuery(pplQuery, previousField))`), or
> 2. Make `stripTimeFilterFromQuery` field-agnostic (match `| where <any identifier> > DATE_SUB(NOW(), INTERVAL ...)` and the legacy `TIMESTAMP(...)` form), since the plugin is the only thing that injects that exact shape.
> 
> Option 2 is simpler and also covers the case where the saved `timestamp_field` is missing or the field got renamed in the index. A unit test on `stripTimeFilterFromQuery` with a mismatched field name would lock it in.

Implemented option 1 as option 2 has a side-effect which removes manually added where clauses like `where <any identifier> > DATE_SUB(NOW(), INTERVAL ...)`

public/components/PplScheduleEditor/PplScheduleEditor.js
vikhy-aws: Fixed

public/components/PplScheduleEditor/PplScheduleEditor.js
vikhy-aws: Fixed

public/components/PplScheduleEditor/PplScheduleEditor.js
vikhy-aws: Fixed

public/pages/CreateMonitor/containers/DefineMonitor/DefineMonitor.js
vikhy-aws: Fixed

riysaxen-amzn:
**Please restore the deleted tests in `pplAlertingHelpers.test.js`**

This file went from 24 tests to 5. The removed suites still target unchanged code and are what's tripping `codecov/project` (-6.07%):

- `computeLookBackMinutes` (5 tests)
- `extractIndicesFromPPL` (5 tests)
- `formatDuration` (4 tests)
- `addTimeFilterToQuery`: 10 of 11 cases, including the legacy `TIMESTAMP(...)` replacement, pipe-position ordering, and the `buildPPLMonitorFromFormik` disable-strips-filter case

The 5 new tests are good and should stay. Suggest appending them to the existing file rather than replacing it. Nothing in the fix requires the old tests to change.

riysaxen-amzn:
Re-reviewed after the rework. The stacked time-filter issue is fixed: I replayed the switch-field and disable-lookback edit paths through `buildPPLMonitorFromFormik` with the persisted `ui_metadata.lookback.timestamp_field` and both now produce a single clause on the current field (or none when disabled). Keeping `stripTimeFilterFromQuery` field-keyed and stripping both previous and current fields is the right call, since it preserves user-written `DATE_SUB` filters on unrelated fields (confirmed with `response_time` case).

Also good: moving the timestamp default out of render into `handlePplDateFieldsChange` via `innerRef`, surfacing a stale saved field as `isInvalid` instead of silently resetting it, and dropping the dead `PplAlertingCreateMonitor` path.

One follow-up worth a tracking issue (not blocking): a PPL monitor created via the API with top-level `timestamp_field` but no `ui_metadata.lookback` will still stack on field switch, because `previousField` resolves to `undefined`. `pplAlertingMonitorToFormik` L47 already knows the original field at hydration time, so stashing it (e.g. `initialTimestampField`) in formik values and falling back to it in `buildPPLMonitorFromFormik` would close that gap.

Will approve once the deleted tests are back.

riysaxen-amzn:
Approving. Re-checked `3721e34` against the version I reviewed this morning: the only delta is `pplAlertingHelpers.test.js`, which is now purely additive (the 24 original tests are restored, the 5 new ones appended, 29 total). All other files are unchanged from `2eb3071`, so the stacked-filter fix, the `innerRef`-based default, and the stale-field `isInvalid` handling are as verified earlier.

On `codecov/project`: that failure is a stale-base comparison, not this PR. Codecov is diffing against `217bf4d`, which is 72 commits behind the actual base (`e1c3ca1`), which is where the +34 files / +2357 lines come from. Patch coverage on the PR's own lines is 100%. Safe to ignore or ask a maintainer to re-run the base upload.

The API-created-monitor edge case I mentioned earlier (no `ui_metadata.lookback`, switch field on edit) stays as a follow-up and doesn't need to block this.

riysaxen-amzn (approved):

lezzago (approved):

PR #1518, opened 2026-09-18
Show data streams in the monitor index picker for local and cross-cluster monitors

vikhy-aws:
### Description
Data streams were missing from the index picker when creating or editing a monitor. The picker resolves indices via `_cat/indices`, which only returns a data stream's backing indices (`.ds-*`), never the stream name itself — so a data stream could never be selected as a monitor's source.

This adds a dedicated `POST /api/alerting/_data_streams route` (backed by `OpensearchService.getDataStreams`, which queries the `/_data_stream/<pattern>` API) and merges the resolved streams into the picker's index options on the client:
  - `MonitorIndex` (standard, non-cross-cluster picker): fetches matching data streams alongside indices/aliases in onFetch and merges them into the exact/partial match lists (deduped by label, sorted).
  - `CrossClusterConfiguration` (cross-cluster picker): resolves the local (hub) cluster's data streams and merges them into that cluster's options, since the remote-indexes API returns only indices/aliases.

Data streams are shaped identically to indices so they flow through the picker as ordinary selectable sources. Notes: the pattern is sanitized (path separators stripped) so it cannot escape the `/_data_stream/` endpoint; a 404 (no matching streams) is treated as an empty result; and remote-cluster data streams are out of scope (merge is gated on the hub cluster).
 
### Issues Resolved
 
### Check List
- [x] All tests pass


public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
```diff
@@ -177,6 +194,20 @@ export class CrossClusterConfiguration extends Component {
...
+        // selectable in the picker.
+        if (clusterInfo.hub_cluster && dataStreams.length) {
+          const dataStreamOptions = dataStreams.map((ds) => ({
+            label: ds.index,
+            health: ds.health,
+            index: ds.index,
+            cluster: clusterInfo.cluster,
+            value: ds.index,
+          }));
+          clusterIndexOptions = _.uniqBy([...clusterIndexOptions, ...dataStreamOptions], 'index');
```
lezzago, line 208: **Correctness (blocking): editing a monitor drops a pre-selected local data stream.**

When editing a monitor whose index list contains a local data stream (`loadedInitialValues === false`), the stream is not in `clusterInfo.indexes`, so the `.map()` above never pushes it to `selectedIndexes`. This merge then adds it to `categorizedIndexOptions[local].options`. The fallback reconciliation below (`Object.entries(indexes).forEach(... if (!includesIndex) selectedIndexes.push(...))`) now sees `includesIndex === true` and skips it.

Net effect: before this change the fallback restored the stream into `selectedIndexes` (it wasn't in options, so `includesIndex` was false); after this change the stream renders as an *unselected* option and is dropped when the edited monitor is saved. Please push merged local data streams into `selectedIndexes` when they appear in the pre-selected `indexes` map, and add a test for the `loadedInitialValues === false` path.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
```diff
@@ -60,8 +60,11 @@ export class CrossClusterConfiguration extends Component {
         dataSourceId: getDataSourceId(),
       };
       const response = await httpClient.get(`../api/alerting/remote/indexes`, { query: query });
+      // The remote-indexes API only returns indices/aliases, so resolve the local
+      // cluster's data streams separately and merge them in during parseOptions.
+      const localDataStreams = await this.getLocalDataStreams();
```
lezzago, line 65: **Efficiency:** `getIndexes()` runs on every `selectedClusters` change (`componentDidUpdate`), and it now unconditionally awaits `getLocalDataStreams()` (always `'*'`) before `parseOptions`. The data-stream list is independent of cluster selection, so every add/remove of a remote cluster re-fetches the entire stream list and serializes ahead of parsing the already-returned remote-indexes response. Consider fetching once (cache) rather than per `getIndexes` call.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
```diff
@@ -71,7 +74,21 @@ export class CrossClusterConfiguration extends Component {
     this.setState({ loading: false });
   }
 
-  parseOptions = (clusterInfos = {}) => {
+  async getLocalDataStreams() {
+    const { httpClient } = this.props;
+    try {
+      const response = await httpClient.post(`../api/alerting/_data_streams`, {
+        body: JSON.stringify({ dataStream: '*' }),
+        query: { dataSourceId: getDataSourceId() },
```
lezzago, line 82: **Consistency:** this builds the query as `{ dataSourceId: getDataSourceId() }`, whereas the rest of the plugin (including `handleQueryDataStreams` in `MonitorIndex.js`) uses `getDataSourceQueryObj()?.query`. With no data source selected this sends `{ dataSourceId: undefined }`, which only works because the http layer drops undefined values. Prefer the shared `getDataSourceQueryObj()` helper for consistency.

public/pages/CreateMonitor/containers/MonitorIndex/MonitorIndex.js
```diff
@@ -171,32 +171,78 @@ class MonitorIndex extends React.Component {
...
+          isLoading: false,
+        });
       });
     } else {
       const partialMatchedIndices = await this.handleQueryIndices(`${query}*`);
       const exactMatchedIndices = await this.handleQueryIndices(query);
       const partialMatchedAliases = await this.handleQueryAliases(`${query}*`);
       const exactMatchedAliases = await this.handleQueryAliases(query);
+      const partialMatchedDataStreams = await this.handleQueryDataStreams(`${query}*`);
+      const exactMatchedDataStreams = await this.handleQueryDataStreams(query);
```
lezzago, line 233: **Efficiency:** the non-wildcard branch now awaits six requests strictly in sequence (2 indices, 2 aliases, 2 data streams) on every search keystroke. These are independent — wrapping them in `Promise.all` would avoid growing picker latency by a full data-stream round-trip per character. (The 4-way serial pattern pre-dates this PR, but the two new awaits extend it.)

lezzago (commented):
Reviewed the data-stream support changes. One functional regression (blocking, see inline on `CrossClusterConfiguration.js`) plus a few robustness/efficiency/consistency notes, and a gap in test coverage.

### Correctness (blocking)
- **Editing a monitor drops a pre-selected local data stream.** See inline comment on the merge block. The new merge makes the data stream appear in the displayed options, which in turn causes the edit-restore fallback to skip adding it to `selectedIndexes` — so saving an edited monitor silently loses the data stream.

### Test coverage gaps
The added tests are good on the happy paths (endpoint shaping, slash-stripping, 404→empty, merge/dedup/sort, `getLocalDataStreams` ok/not-ok/throws), but they miss the one path that's actually broken and a few edges:
1. **No test for the edit-restore path** (`loadedInitialValues === false` with a data stream in `values.index`). Every `parseOptions` merge test asserts only on the displayed `indexOptions`, never on `selectedIndexes` during an edit — exactly the regression above. A test there would fail today.
2. No assertion that client-side `health` is surfaced correctly on merged streams (server test covers `GREEN`→`green`; client passes `ds.health` through untested).
3. `handleQueryDataStreams` error / not-ok branches are untested in `MonitorIndex` (only happy path + `*:` guard).
4. The empty-string (`''`) guard in `handleQueryDataStreams` is untested (only `*:`).
5. Server: the non-404 error branch (`ok: false, resp: err.message`) and the `data_streams` missing/undefined fallback are untested.

Overall the change is well-structured and mostly well-tested; the edit-restore regression (#1) is the item worth blocking on.

server/services/OpensearchService.js
```diff
@@ -111,6 +111,50 @@ export default class OpensearchService extends MDSEnabledClientService {
...
+      // wildcard ('*') intact while preventing the pattern from escaping the
+      // /_data_stream/ path.
+      const pattern = String(dataStream || '*').replace(/[/\\]/g, '') || '*';
+      const client = await this.getClientBasedOnDataSource(context, req);
+      // _cat/indices (used by getIndices) only returns a data stream's backing
+      // indices (.ds-*), never the stream name, so data streams are resolved
+      // separately here and merged into the index picker on the client.
+      const response = await client('transport.request', {
+        method: 'GET',
+        path: `/_data_stream/${pattern}`,
```
lezzago, line 130: **Robustness:** `pattern` is interpolated into the path with only `/` and `\` stripped, not URL-encoded. A pattern containing a space, `%`, comma, etc. is passed through raw and can produce a malformed path or an unintended multi-target request. Slash-stripping blocks endpoint escape but not general URL safety — consider `encodeURIComponent(pattern)` on the sanitized value.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
vikhy-aws: Fixed

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
vikhy-aws: Fixed

server/services/OpensearchService.js
vikhy-aws: This matches the existing `index/alias` pattern.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
```diff
@@ -71,7 +74,30 @@ export class CrossClusterConfiguration extends Component {
     this.setState({ loading: false });
   }
 
-  parseOptions = (clusterInfos = {}) => {
+  async getLocalDataStreams() {
+    // The data stream list is independent of cluster selection, so resolve it once
+    // and reuse it — getIndexes() runs on every selectedClusters change and would
+    // otherwise re-fetch the entire list on each add/remove of a remote cluster.
+    if (this.localDataStreams) return this.localDataStreams;
```
lezzago, line 81: **Stale data-stream cache under multi-data-source (MDS).**

`this.localDataStreams` is cached but never keyed by or invalidated on `dataSourceId`. Under MDS, if a user opens a cross-cluster monitor (caching data source A's streams), switches the data source selector to B, then adds/removes a remote cluster, `getIndexes()` re-runs: the remote indexes reflect B (because `getDataSourceId()` is re-evaluated per call) but `getLocalDataStreams()` returns the cached A stream list.

Unlike `MonitorIndex` — which re-fetches on `isDataSourceChanged` in `componentDidUpdate` — this component has no data-source-change handling and isn't remounted by the parent, so the stale list persists and a data source A stream can be selected and saved against B.

Suggest keying the cache on the current data source id (invalidate/refetch when it changes), or clearing `this.localDataStreams` on data-source change.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
lezzago, line 65: **First-call serialization.** On mount the remote-indexes GET (line 62) and this data-streams POST are independent, but the `await` on 62 fully completes before the POST is even issued — adding a full data-stream round-trip to initial picker latency. The added cache addresses *repeat* calls (the reviewer's concern) but not this first call. Wrapping the two independent requests in `Promise.all` would remove the extra round-trip on mount.

public/pages/CreateMonitor/components/CrossClusterConfigurations/containers/CrossClusterConfiguration.js
```diff
@@ -177,6 +203,36 @@ export class CrossClusterConfiguration extends Component {
...
+            // are absent from clusterInfo.indexes, so unlike regular indexes they are
+            // not captured by the .map() above. Because this merge adds them to the
+            // displayed options, the reconciliation below would see them as "already
+            // an option" and skip restoring them into selectedIndexes — silently
+            // dropping the stream on save. Push them here instead.
+            if (
+              !loadedInitialValues &&
+              (indexes[clusterName] || []).includes(dataStreamOption.index)
+            ) {
+              selectedIndexes.push(dataStreamOption);
```
lezzago, line 229: **Possible duplicate in `selectedIndexes` during edit (edge case).** If a pre-selected local name appears both as a regular index (pushed by the `.map()` above) and as a data stream, it is pushed here a second time, and `renderSelectedClusterIndexesOptions` only maps — it doesn't dedup — so the duplicate survives into the form value and combo box. Low likelihood in practice since a stream's backing indices are `.ds-*` (the stream name rarely also surfaces as a plain index), but a `_.uniqBy` on the final `selectedIndexes`, or a guard before this push, would close it.

eirsep (approved):

lezzago (approved):

